PR #351, opened 2014-12-18
Memcache prep

TheDudeWithTheThing:
Get that CACHE playa

:moneybag: :moneybag:  :moneybag:  :moneybag:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -79,7 +79,7 @@ group :test do
 end
 
 group :development do
-  gem 'capistrano', '~> 3.1.0'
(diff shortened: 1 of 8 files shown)


docvego:
Yeah, keep it simple for now and the slow pages aren't a serious issue at least not yet.

docvego:
For future reference,  Caching caused problems with the feature of linking dogs to applications via the Adopters show screen.  That bit was backed out in 7c2406ee6b5c0e5ff6e55ac3e0db0699b1e9afc5  

We'll revisit caching later possibly as part of the Rails 4 migration.

PR #353, opened 2014-12-22
WIP making comments editable

docvego:
Not ready for merging.   @TheDudeWithTheThing 

Save comment doesn't currently work, and Javascript can be cleaned up to avoid the need for the IDs.


docvego:
Okay, so the Button labeled save in the _comments.html.erb partial was rendered as type=submit so the save was being double requested hence all my total confusion as to what was going on.   

Current main issue is the POST is trying to go to the adopters controller instead of the comments controller, and hitting a 404 Error.  It's also trying to do it a whole bunch of times.

app/assets/javascripts/comments.js
```diff
@@ -1,5 +1,46 @@
...
+          $parent.find('.read-only-comment').hide();
+          $parent.find('.editable-comment').show();
+          $parent.find('.toggle-edit-comment').removeClass('btn-primary').text('Cancel');
+          $parent.find('.save-edit-comment').show();
+        }
+        isEditing = !isEditing;
+
+      });
+
+      $('.save-edit-comment').click(function() {
```
TheDudeWithTheThing: this whole block is being executed for every comment, so when you click a save button it will be executed comment # of times, you'll need to move it out and as a result the $parent.find stuff should get their own functions so you can shut off edit mode from the save function.

```javascript
function editComment($parent) {
  $parent.find('.read-only-comment').hide();
  $parent.find('.editable-comment').show();
  $parent.find('.toggle-edit-comment').removeClass('btn-primary').text('Cancel');
  $parent.find('.save-edit-comment').show();
}
```

app/assets/javascripts/comments.js
```diff
@@ -1,5 +1,46 @@
...
+        var $parent = $(event.target).parents('.edit-comment')
```
TheDudeWithTheThing: this is finding a div and not the form, change to

```javascript
var $parent = $(event.target).parents('form.edit_comment')
```

app/assets/javascripts/comments.js
```diff
@@ -43,6 +84,20 @@ $( function () {
   });
 });
 
+function save_comment($parent) {
```
TheDudeWithTheThing: form will be passed in now using the previous change so this should work ok

app/controllers/comments_controller.rb
```diff
@@ -36,7 +37,8 @@ def update
     @comment = Comment.find(params[:id])
     if @comment.user_id == current_user.id
       @comment.update_attributes(params[:comment])
-      respond_with @comment
+      respond_with @comment.content
+      end
```
TheDudeWithTheThing: random `end` here causing a 500

docvego:
Awesome thanks! Will be making these changes tonight

docvego:
FINALLY okay please review and test and stuff and let me know how it goes.  

Only minor issues I'm aware of:
1) Clicking cancel doesn't reset the textbox content
2) Have to hit edit twice after a save

app/assets/javascripts/comments.js
```diff
@@ -43,6 +67,38 @@ $( function () {
...
+function showComment($parent) {
+  $parent.find('.read-only-comment').show();
+  $parent.find('.editable-comment').hide();
+  $parent.find('.toggle-edit-comment').addClass('btn-primary').text('Edit');
+  $parent.find('.save-edit-comment').hide();
+}
+
+
+function saveComment($form) {
+  var $url = $form.attr('action');
```
TheDudeWithTheThing: I use $vars for jquery objects, no need to use for `$url`

TheDudeWithTheThing:
pushed a fix for the 2nd problem, nice and easy

docvego:
Thanks!  Seeing two other problems.

3) Comment edits don't work at all in the dogs page (probably a problem with the DOM, working on this.)
4) After posting a new comment the edit buttons don't work at all.  (Any guesses on this one?)

TheDudeWithTheThing:
> 3) Comment edits don't work at all in the dogs page (probably a problem with the DOM, working on this.)
yeah I think the comment form is named differently there or something, check the "ghetto namespacing"

app/assets/javascripts/comments.js
```diff
@@ -1,5 +1,31 @@
 $( function () {
-  $('.edit_comment #comment_content').live('change', RescueRails.saveParentForm);
+
+  if ( $('.edit_comment').length > 0) {
+
+    $('.toggle-edit-comment').each(function(){
```
TheDudeWithTheThing: > 4) After posting a new comment the edit buttons don't work at all. (Any guesses on this one?

you'll want to make this line `.on('click', function() {` instead of `.each` and then you'd have to remove the isEditing and click lines.  Then instead of an isEditing var maybe store it in a data attribute somewhere?

docvego:
Okay, problem 3) is taken care of.   

Not sure about how to toss isEditing somewhere else.  Does that go into the DOM somehow?  Is this barking up the right tree?  http://api.jquery.com/data/

docvego:
Looks like that's the ticket.  Just about have it hacked together for review.

docvego:
Stuck

app/assets/javascripts/comments.js
```diff
@@ -1,5 +1,29 @@
 $( function () {
-  $('.edit_comment #comment_content').live('change', RescueRails.saveParentForm);
+
+  if ( $('.edit_comment').length > 0) {
+
+    $('.toggle-edit-comment').click(function(){
```
TheDudeWithTheThing: k got it (these are the things you don't worry about and forget once you're doing ember =\)

```javascript
$(document).on('click', '.toggle-edit-commit', function(){
```

docvego:
Looks like we aren't quite there yet.  I'm unable to get the edit button to work in IE8, IE9 or IE10

Works okay in Chrome and IE11.

TheDudeWithTheThing:
Oh no, I don't think they support data attributes

docvego:
It looks like they're supposed to, I mean not even the current version of Firefox?    Something ain't right.

TheDudeWithTheThing:
nice, was going to change it before but it was working somehow so I left it alone, see code!

app/assets/javascripts/comments.js
```diff
@@ -1,5 +1,29 @@
 $( function () {
-  $('.edit_comment #comment_content').live('change', RescueRails.saveParentForm);
+
+  if ( $('.edit_comment').length > 0) {
+
+    $(document).on('click', '.toggle-edit-comment', function(){
+      var $parent = $(event.target).parents('form.edit_comment');
```
TheDudeWithTheThing: event is a magic var in chrome or something, get that out!  the event is passed into the function, so

```
 function(e) {
 $(e.target)
````

docvego:
Okay, I think this is all set.   Let me know if there are any other suggested changes!

PR #360, opened 2015-02-06
Fix/controller specs

docvego:
Taking a look at the controller and model spec stuff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ config/database.yml
 chromedriver.log
 .idea/
 .env.*
+coverage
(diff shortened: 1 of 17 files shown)


docvego:
So for the test of having a valid factory, it looks like we have a few options:

`expect(create(:banned_adopter)).to be_valid`
`expect(build(:banned_adopter)).to be_valid`
`expect(banned_adopter).to be_valid`

`expect(banned_adopter.valid?).to eq(true)`

TheDudeWithTheThing:
DO IT, this can be our baseline and we can add more tests in other branches

:+1:

PR #362, opened 2015-02-12
Travis CI and Coveralls Setup / Controller Spec Cleanup

docvego:
This gets travis CI and coveralls rolling.  Now just need to get all those controller tests to pass!


docvego:
Okay, goal for this branch is to get the existing controller specs updated to the newer rspec syntax, get them to all pass, and update any other depreciation warnings.

PR #367, opened 2015-02-19
DogsController#index tests

TheDudeWithTheThing:
* Adds json_spec for rspec json helpers
* Configures JSON helpers
* Adds tests for index method, all if / else paths

First Commit, add tests: https://github.com/ophrescue/RescueRails/commit/d17773dcd469241f898b927f7d5fed0d75520203

### DIFF
--- a/app/controllers/dogs_controller.rb
+++ b/app/controllers/dogs_controller.rb
@@ -1,13 +1,13 @@
 class DogsController < ApplicationController
-  helper_method :sort_column, :sort_direction, :is_fostering_dog?
+  helper_method :is_fostering_dog?
 
-  autocomplete :breed, :name, :full => true
+  autocomplete :breed, :name, full: true
 
-  before_filter :authenticate,                            :except => [:index, :show]
-  before_filter :edit_dog_check,                          :only => [:edit, :update]    
-  # before_filter :fostering_dog_user, :edit_dogs_user,     :only => [:edit, :update]                                   
-  before_filter :add_dogs_user,                          :only => [:new, :create]
-  before_filter :admin_user,                              :only => [:destroy]
+  before_filter :authenticate, except: %i(index show)
+  before_filter :edit_dog_check, only: %i(edit update)
+  before_filter :add_dogs_user, only: %i(new create)
+  before_filter :admin_user, only: %i(destroy)
+  before_filter :load_dog, only: %i(show edit update destroy)
 
   def index
     is_manager = signed_in? && session[:mgr_view]
@@ -22,7 +22,6 @@ class DogsController < ApplicationController
   end
 
   def show
-    @dog = Dog.find(params[:id])
     sort_dog_photos
     @title = @dog.name
   end
@@ -34,7 +33,6 @@ class DogsController < ApplicationController
   end
 
   def edit
-    @dog = Dog.find(params[:id]) 
     @dog.primary_breed_name = @dog.primary_breed.name unless @dog.primary_breed.nil?
     @dog.secondary_breed_name = @dog.secondary_breed.name unless @dog.secondary_breed.nil?
     sort_dog_photos
@@ -43,7 +41,6 @@ class DogsController < ApplicationController
   end
 
   def update
-    @dog = Dog.find(params[:id])
     if @dog.update_attributes(params[:dog])
       flash[:success] = "Dog updated."
       redirect_to @dog
@@ -71,7 +68,7 @@ class DogsController < ApplicationController
   end
 
   def destroy
-    Dog.find(params[:id]).destroy
+    @dog.destroy
     flash[:success] = "Dog deleted."
     redirect_to dogs_path
   end
@@ -88,6 +85,10 @@ class DogsController < ApplicationController
 
 
   private
+
+    def load_dog
+      @dog = Dog.find(params[:id])
+    end
 
     def sort_dog_photos
       @dog.photos.sort_by!{|photo| photo.position}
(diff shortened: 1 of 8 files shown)


TheDudeWithTheThing:
whoa, thanks coveralls!

spec/controllers/dogs_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+        it_behaves_like 'an OK response with 1 dog'
+      end
+
+      context 'search by name' do
+        let!(:found_dog) { create(:dog, name: 'oscar') }
+        let(:params) { { search: 'oscar' } }
+
+        before do
+          get_index
+        end
```
jayroh: since this is only one small short statement you could roll this block up into one inline proc. General pattern I've seen is - more than one line? do/end. One line statement, `{}`s are fine. Small coding style nitpick.

spec/controllers/dogs_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    it 'contains found dog' do
+      expect([found_dog].to_json).to be_json_eql(response.body)
+    end
+  end
+
+  describe 'GET #index' do
+    context 'user is logged in' do
+      let!(:found_dog) { create(:dog, tracking_id: 1) }
+      let!(:other_dog) { create(:dog, tracking_id: 100) }
```
jayroh: would a `build_stubbed` strategy (instead of `create`) still work here? Would save a trip and transaction or two to the db.

app/models/dog_searcher.rb
```diff
@@ -0,0 +1,76 @@
...
+      else
+        @dogs = Dog.where("name ilike ?", "%#{@params[:q]}%")
+      end
+    else
+      @dogs = Dog.where("status IN (?)", PUBLIC_STATUSES)
+    end
+
+    includes
+    sort
+    paginate
```
jayroh: these previous three methods could be named with a little more intent.  Maybe ..

```
@dogs = with_eager_load_tables
@dogs = with_assigned_sort_order
@dogs = for_page(@params[:page])
```

¯\_(ツ)_/¯

app/controllers/dogs_controller.rb
```diff
@@ -10,26 +10,10 @@ class DogsController < ApplicationController
...
-      end
-    else
-      @title = "Our Dogs"
-      statuses = ['adoptable', 'adoption pending', 'coming soon']
-      @dogs = Dog.where("status IN (?)", statuses).order(sort_column + ' ' + sort_direction).paginate(:per_page => 30, :page => params[:page]).includes(:photos, :primary_breed)
-    end
-    ## Need to support the dog select drop down in the adopt app as well!
+    is_manager = signed_in? && session[:mgr_view]
+
+    @title = is_manager ? "Dog Manager" : "Our Dogs"
```
jayroh: Does the concept of a current user exist in this controller yet?

If so, instead of having to set a `@title` ivar in the controller you could spit out the appropriate string directly from the user object (if it's available).

`<%= current_user.title %>`

Again, incremental is key here so I'm not suggesting the change be made in this PR, but I'm curious if this is something I could help extract / simplify.

TheDudeWithTheThing: title as in job title and not page title right?  The problem is knowing which view you're in since you can switch between a normal view and an admin / manager view

app/models/dog_searcher.rb
TheDudeWithTheThing: fully agree, I was just blasting this out on the train, I suck at names

spec/controllers/dogs_controller_spec.rb
TheDudeWithTheThing: well it's still has to search / hit the DB, or you mean once these tests are moved out to the searcher spec we stub the search returning something to satisfy the index returning data?

app/controllers/dogs_controller.rb
jayroh: If this is a page title then that Would make me ask -

1. Should this be split into two different controllers/resources? If it's doubling up on responsibility then I'd say yes. It should. 

Or 2. If not - why not make the title something more generic? Setting a page title in the controller is a smell - total view concern.

spec/controllers/dogs_controller_spec.rb
jayroh: Ah duh that's a good point. Wasn't entirely obvious to me when reading this. 

Which is why I like my 4 phases of a test as close to each other as possible :D

TheDudeWithTheThing: actually I agree with you here, that should be totally generic and then overwritten in the search by tracking_id context, will change when I move these tests to dog_searcher_spec

docvego:
Page title could totally be set in the views.  

We basically have a need for two different views of dogs#index:

The simple one that's basically a photo album of available dogs and the "boring ugly spreadsheet" "Manager View". 

Setting title in controllers was leftover from the old Rails 3.x Hartl tutorial when I went through that.

app/controllers/dogs_controller.rb
docvego: Page title could totally be set in the views.

We basically have a need for two different views of dogs#index:

The simple one that's basically a photo album of available dogs and the "boring ugly spreadsheet" "Manager View".

Setting title in controllers was leftover from the old Rails 3.x Hartl tutorial when I went through that.

TheDudeWithTheThing:
@docvego @jayroh I think I'm done here for now, not too bad now!

docvego:
Damn dude, this looks great.  Want to make sure I understand how everything works and then I'll be merging it on in.

TheDudeWithTheThing:
:stuck_out_tongue_winking_eye: :dancers:

docvego:
Also @jayroh let me know if you have any other thoughts before I merge dis in.

PR #375, opened 2015-03-05
Adopter controller clean up

TheDudeWithTheThing:
Adopter controller gets the dog controller treatment.

* Move index search code to AdopterSearcher
* AdopterSearcher spec
* Create load_adopter before_filter

### DIFF
--- a/app/controllers/adopters_controller.rb
+++ b/app/controllers/adopters_controller.rb
@@ -82,7 +82,7 @@ class AdoptersController < ApplicationController
 
   private
 
-  def laod_adopter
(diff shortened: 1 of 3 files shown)


app/controllers/adopters_controller.rb
```diff
@@ -97,6 +82,10 @@ def check_email
 
   private
 
+  def laod_adopter
```
docvego: Le Typo

TheDudeWithTheThing:
I hadn't even run the code yet but just did, all is well!

PR #378, opened 2015-03-05
adds validation and tests for user state and zip code

docvego:
Fairly simple stuff, check it out @TheDudeWithTheThing

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -12,12 +12,10 @@ describe User do
     context 'with valid fields' do
       it 'should accept a two letter state' do
         user = build(:user, state: 'PA')
-        user.valid?
         expect(user).to be_valid
       end
       it 'should accet a 5 digit zipcode' do
-        user =build(:user, zip: '12345')
-        user.valid?
+        user = build(:user, zip: '12345')
         expect(user).to be_valid
       end
     end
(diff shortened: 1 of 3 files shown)


spec/models/user_spec.rb
```diff
@@ -6,6 +6,37 @@
...
   end
 
+
+  describe 'contact information' do
+
+    context 'with valid fields' do
+      it 'should accept a two letter state' do
+        user = build(:user, state: 'PA')
+        user.valid?
+        expect(user).to be_valid
```
TheDudeWithTheThing: so what `be_` does is takes whatever you put after it and calls xxx? on it, so if your model had a method called `horsey?` you could do

```
expect(user).to be_horsey
```

and it would get called.  

Short answer is you shouldn't need that first `user.valid?`

spec/models/user_spec.rb
```diff
@@ -6,6 +6,37 @@
...
+      end
+      it 'should accet a 5 digit zipcode' do
+        user =build(:user, zip: '12345')
```
TheDudeWithTheThing: space here

spec/models/user_spec.rb
```diff
@@ -6,6 +6,37 @@
...
+        user.valid?
```
TheDudeWithTheThing: no need for `user.valid?` here 

*see above*

PR #383, opened 2015-03-09
Foster Mentor Tracking Resolves #234

### LINKED ISSUE #234: Foster Mentor Tracking
"Can we include a slot for a foster mentor on our foster user page?  We want to be able to assign a foster mentor to fosters and be able to see who that person is in dms.  This will help the foster coordinators in tracking."   -JD

docvego:


TheDudeWithTheThing:
If you're a mentor you can't also be a mentee then right?  Just wondering about the double use of mentor_id

app/assets/javascripts/chosen-calls.js
```diff
@@ -6,4 +6,5 @@ jQuery(function($){
 	$('#dog_foster_id').chosen();
 	$('#dog_coordinator_id').chosen();
 	$('#dog_shelter_id').chosen();
+  $('#user_mentor_id').chosen();
```
TheDudeWithTheThing: might as well fix the previous lines spacing too, unless github is being weird with displaying

docvego:
Correct, you aren't allowed to mentor yourself. However I don't see a need to actually enforce that programmatically.

TheDudeWithTheThing:
yup got it, looks good!

:+1:

PR #386, opened 2015-03-10
resolves #384 error creating comments when not signed in

### LINKED ISSUE #384: Error Creating Comments when not signed in
Trying to create a comment when a user's session has expired results in the following in the logs:

        A NoMethodError occurred in comments#create:
        undefined method `id' for nil:NilClass
        app/controllers/comments_controller.rb:23:in `create'

Use gets an "Error Saving Comment" message, and usually just keeps hitting the button and getting annoyed.

docvego:
User will now get prompted to reenter credentials.
However it renders in the past comments section instead of
refreshing the page.  Dont thinks its worth ripping the jquery apart to make this elegent.

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -1,4 +1,7 @@
 class CommentsController < ApplicationController
+  include SessionsHelper
+
+  before_filter :authenticate
 
   respond_to :html, :json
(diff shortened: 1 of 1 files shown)


TheDudeWithTheThing:
keep it simple

Issue #409, opened 2015-09-12
Implement Strong_parameters


markottaviani:
https://github.com/rails/strong_parameters

Doing this as prep for rails4

PR #410, opened 2015-09-13
Strong Paramaters for User Model

markottaviani:
Work Completed on User Model & Ready to Merge :  In support of #409 

My first attempt at setting up strong parameters for the User model/controller.  

Please check this out and make sure I'm moving in the right direction before I attack the rest of the application.  

Will need to remove the commented out code still too.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -52,7 +52,7 @@ class UsersController < ApplicationController
 
   def update
     @user.accessible = :all if current_user.admin?
-    if @user.update_attributes(params[:user])
+    if @user.update_attributes(user_params)
       @user.update_attribute(:lastverified, Time.now)
       flash[:success] = "Profile updated."
       redirect_to @user
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -25,16 +25,13 @@ class UsersController < ApplicationController
   end
 
   def new
-    @user = User.new(user_params)
+    @user = User.new
     @title = "Add a Staff Account"
     init_fields
   end
 
   def create
     @user = User.new(user_params)
-    # @user.accessible = :all if current_user.admin?
-    @user.attributes = params[:user]
-    @user.email.downcase!
     if @user.save
       flash[:success] = "Account created for " + @user.name
       redirect_to users_path
@@ -54,7 +51,6 @@ class UsersController < ApplicationController
   end
 
   def update
-    params[:user][:email].downcase!
     @user.accessible = :all if current_user.admin?
     if @user.update_attributes(params[:user])
       @user.update_attribute(:lastverified, Time.now)
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -51,7 +51,6 @@ class UsersController < ApplicationController
   end
 
   def update
-    @user.accessible = :all if current_user.admin?
     if @user.update_attributes(user_params)
       @user.update_attribute(:lastverified, Time.now)
       flash[:success] = "Profile updated."
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -25,14 +25,14 @@ class UsersController < ApplicationController
   end
 
   def new
-    @user = User.new
+    @user = User.new(user_params)
     @title = "Add a Staff Account"
     init_fields
   end
 
   def create
-    @user = User.new
-    @user.accessible = :all if current_user.admin?
+    @user = User.new(user_params)
+    # @user.accessible = :all if current_user.admin?
     @user.attributes = params[:user]
     @user.email.downcase!
     if @user.save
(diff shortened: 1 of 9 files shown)


markottaviani:
I think I need to add some tests and make some modifications for User controllers Update method.  As I'm guessing that won't actually work as-is

app/controllers/users_controller.rb
```diff
@@ -75,6 +75,39 @@ def destroy
 
   private
 
+    def user_params
+      if current_user && current_user.admin?
+        params.require(:user).permit!
```
TheDudeWithTheThing: Does this say allow everything for admins?  I think you'd still want to restrict what params are permitted since you may have malicious admins!

app/controllers/users_controller.rb
```diff
@@ -25,14 +25,14 @@ def show
   end
 
   def new
-    @user = User.new
+    @user = User.new(user_params)
```
TheDudeWithTheThing: no need for `user_params` here

app/controllers/users_controller.rb
```diff
@@ -25,14 +25,14 @@ def show
...
     @title = "Add a Staff Account"
     init_fields
   end
 
   def create
-    @user = User.new
-    @user.accessible = :all if current_user.admin?
+    @user = User.new(user_params)
+    # @user.accessible = :all if current_user.admin?
     @user.attributes = params[:user]
```
TheDudeWithTheThing: this line can hit the bricks, your attributes are already set from `@user = User.new(user_params)`

app/controllers/users_controller.rb
```diff
@@ -25,14 +25,14 @@ def show
...
     @user.email.downcase!
```
TheDudeWithTheThing: I'd make this a `before_save` on the user model

TheDudeWithTheThing:
> I think I need to add some tests and make some modifications for User controllers Update method. As I'm guessing that won't actually work as-is

correct, the biggest change is

```ruby
if @user.update_attributes(user_params)
```

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    context 'logged in as an admin' do
+      before :each do
+        allow(controller).to receive(:current_user) { admin }
+      end
+
+      it 'is able to create a user' do
+        expect{
+          post :create, user: attributes_for(:user)
+        }.to change(User, :count).by(2)
```
TheDudeWithTheThing: why 2??

TheDudeWithTheThing: oh is it because creating an admin counts towards the `User.count`?  I hate when it does that.

TheDudeWithTheThing:
k I'm done looking for now, let me know questions / comments / when to look again...

spec/controllers/users_controller_spec.rb
markottaviani: Good question! I have no friggan idea. My only guess is that it's counting the admin account? I'm doing something wrong.

TheDudeWithTheThing: nah I think that's it, maybe change the test to `User.admin(false).count` ?

app/controllers/users_controller.rb
markottaviani: True, I could whitelist the specific admin tier variables, but at this point an admin can modify any user param from the web interface anyway, so the results would be same.

markottaviani:
Thanks, I'll let you know

spec/controllers/users_controller_spec.rb
markottaviani: Ah ha:

*`let` is evaluated lazily. That means that the code inside the block isn't run until the variable is used. Since the job variable is used for the first time within the expect blocks, the job count increases. If you want the code inside the let blocks to be run immediately, use the `let!` method.*

markottaviani:
Okay, made the suggested clean ups, only rub is figuring out how the heck to write these PUT update tests.  Added some pseudocode

markottaviani:
Think I found the reference I needed.   http://everydayrails.com/2012/04/07/testing-series-rspec-controllers.html  making test progress

markottaviani:
Okay, ready for another review!

app/controllers/users_controller.rb
```diff
@@ -54,9 +51,8 @@ def edit
   end
 
   def update
-    params[:user][:email].downcase!
     @user.accessible = :all if current_user.admin?
```
TheDudeWithTheThing: does strong_params make it so you don't need this anymore?  not sure what this is for

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    before :each do
+      @test_user = create(:user)
+    end
+
+    context 'logged in as admin' do
+      before :each do
+        allow(controller).to receive(:current_user) { admin }
+      end
+
+      it 'locates the requested user' do
```
TheDudeWithTheThing: this is pretty basic, I don't think I've written tests for this before

spec/controllers/users_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        expect{
+          post :create, user: attributes_for(:user)
+        }.to change(User, :count).by(0)
+      end
+
+    end
+
+  end
+
+  describe 'PUT update' do
```
TheDudeWithTheThing: good stuff!  I've cleaned it up a bit

```ruby
  describe 'PUT update' do
    let(:test_user) { create(:user, admin: FALSE) }
    let(:request) { -> {put :update, id: test_user.id, user: attributes_for(:user, admin: TRUE)} }

    context 'logged in as admin' do
      before :each do
        allow(controller).to receive(:current_user) { admin }
      end

      it 'updates the users permissions' do
        expect { request.call }.to change{ test_user.reload.admin}.from(FALSE).to(TRUE)
      end

    end

    context 'logged in as normal user' do
      before :each do
        allow(controller).to receive(:current_user) { hacker }
      end

      it 'is unable to modify user permissions' do
        expect { request.call }.to_not change{ test_user.reload.admin}
      end
    end
  end
```

TheDudeWithTheThing:
Looking good, this is a great start to getting ready for rails 4.

app/controllers/users_controller.rb
markottaviani: yeah, good catch.  Removing

markottaviani:
okay, changes applied.  Good to Merge?

TheDudeWithTheThing:
I think we're good!

:+1:

PR #414, opened 2015-10-17
WIP- Strong attributes for dogs

markottaviani:
Trying to add factories for the nested attributes of the dog, attachments & photos.  

* Failures:

  1) DogsController GET #index user is logged in returns dog as json
     Failure/Error: let!(:dog) { create(:dog) }
     NoMethodError:
       undefined method `attachment=' for #<Dog:0x0000010e290c48>


markottaviani:
@TheDudeWithTheThing @jayroh any insight welcome.

TheDudeWithTheThing:
I don't think you can use association with a has_many relationship.  you might have to make a trait that adds to the attachments var after the dog is created.

markottaviani:
Okay, I think this is good to go.

TheDudeWithTheThing:
awesome, nothing glaring on my end, everything still works?

:+1:

markottaviani:
haha I think so.  I might run it through the staging server to make sure there isn't any S3 weirdness

markottaviani:
spoke too soon two of these tests fail when trying ot use the :dog_with_photo_and_attachment factory. Also, unable to add photos or attachments to a dog after it is created when manually testing

TheDudeWithTheThing:
yeah that was annoying when I originally wrote tests for it, I don't even think there are docs on how to test that stuff either

markottaviani:
Correction, unable to edit any fields for an existing dog.  

`1 error prohibited this dog from being saved:
There were problems with the following fields:

Attachments attachment can't be blank`

markottaviani:
Think I got a lead

`Unpermitted parameters: _destroy, id`

http://stackoverflow.com/questions/17686440/deletion-of-nested-attributes-does-not-work

markottaviani:
Rolled the testes back to be lest complicated and passing.  Still better off than where we were.  Merging and deploying

PR #429, opened 2016-01-09
Folders Refactoring

markottaviani:
Quick refactoring of the Folders/Files System.  Folders are now clicked on individually instead of displaying the entire contents of every single folder all the time!


TheDudeWithTheThing:
looks like you just broke up the template to the proper show action instead of showing all data on one page?  If that's it then this is good to go!

:+1:

PR #439, opened 2016-03-05
fix for nil max hrs alone

markottaviani:
@TheDudeWithTheThing is there a better control structure to do for this?  Or is this a good choice?

### DIFF
--- a/spec/features/view_adoption_app_spec.rb
+++ b/spec/features/view_adoption_app_spec.rb
@@ -1,0 +1,17 @@
+require 'rails_helper'
+
+feature 'View a mostly nil Adoption App without crashing', js: true do
+  let!(:admin) {create(:user, :admin)}
+  let!(:adopter_with_null_app) {create(:adopter_with_null_app)}
+
+  scenario "Adoptor barely fills out adoption application" do
+
+    sign_in(admin)
+
+    visit '/adopters'
+    expect(page).to have_content('Adoption Applications')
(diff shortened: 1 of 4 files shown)


TheDudeWithTheThing:
Use a view helper to take the logic out of the template 

http://mixandgo.com/blog/the-beginner-s-guide-to-rails-helpers

TheDudeWithTheThing:
Then that becomes 1 span which calls max_hours_alone_css_class(max_hrs_alone) in the class= 

Probably can have a better name than that?

markottaviani:
Added some tests, merging and deploying this so the problem is fixed, going to refactor with helpers in a fresh branch.

spec/features/view_adoption_app_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    sign_in(admin)
+
+    visit '/adopters'
+    expect(page).to have_content('Adoption Applications')
+    click_link(adopter_with_null_app.name)
+    expect(page).to have_content(adopter_with_null_app.name)
+
+  end
+end
```
TheDudeWithTheThing: Nooooo!

What editor? Make it always save with a beeline as the last char

markottaviani: using Sublime Text 2, ah ah, history lesson found: https://robots.thoughtbot.com/no-newline-at-end-of-file

PR #440, opened 2016-03-06
refactor max hrs alone to use a view helper

markottaviani:
@TheDudeWithTheThing thanks for the advice, how's this look?

### DIFF
--- a/app/helpers/adopters_helper.rb
+++ b/app/helpers/adopters_helper.rb
@@ -1,0 +1,39 @@
+# == Schema Information
+#
+# Table name: adopters
+#
+#  id                  :integer          not null, primary key
+#  name                :string(255)
+#  email               :string(255)
+#  phone               :string(255)
+#  address1            :string(255)
+#  address2            :string(255)
+#  city                :string(255)
+#  state               :string(255)
+#  zip                 :string(255)
+#  status              :string(255)
+#  when_to_call        :string(255)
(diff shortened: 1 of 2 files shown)


markottaviani:
*bump* :smile:

TheDudeWithTheThing:
oh oops, when did you do this?

app/helpers/adopters_helper.rb
```diff
@@ -0,0 +1,39 @@
...
+#  flag                :string(255)
+#  is_subscribed       :boolean          default(FALSE)
+#  completed_date      :date
+#
+
+module AdoptersHelper
+  def max_hours_alone_label(hrs_alone)
+    case
+    when hrs_alone.nil?
+      raw("<span class='label label-warning'>?? hrs alone</span>")
```
TheDudeWithTheThing: so much repeating of html!

why not make a function to return the class and another to return the "decorated" `hrs_alone` value?

@jayroh has extensive knowledge here, what say you?

TheDudeWithTheThing:
ah make sure you add new comments so I get notified, not just change the description  :smile_cat:

app/helpers/adopters_helper.rb
markottaviani: Good idea, though I think I'm going to kick the can down the road on this since this is all going to get torn apart whenever i get around to rebuilding the front-end ui.

PR #444, opened 2016-03-26
DONE - Add Confidentiality Agreement

markottaviani:
Okay, I can't figure out what I'm doing wrong here...

### DIFF
--- a/app/views/users/_form.html.erb
+++ b/app/views/users/_form.html.erb
@@ -97,11 +97,11 @@
           <%= f.label :agreement, 'Foster Agreement', :class => 'control-label' %>
           <div class="controls">
           <%= f.fields_for :agreement do |af| %>
-            <% if af.object.new_record? %>
+            <% if af.object.persisted? %>
               <%= af.file_field :attachment %>
-            <% else %>
-                <span>Foster Agreement on File, overwrite?</span> <%= af.file_field :attachment %>
+                <span>Foster Agreement on File, overwrite?</span>
             <% end %>
+            <%= af.file_field :attachment %>
           <% end %>
           </div>
         </div>
@@ -112,11 +112,10 @@
           <%= f.label :confidentiality_agreement, 'Confidentiality Agreement', :class => 'control-label' %>
           <div class="controls">
           <%= f.fields_for :confidentiality_agreement do |af| %>
-            <% if af.object.new_record? %>
-              <%= af.file_field :attachment %>
-            <% else %>
-                <span>Confidentiality Agreement on File, overwrite?</span> <%= af.file_field :attachment %>
+            <% if af.object.persisted? %>
+              <span>Confidentiality Agreement on File, overwrite?</span>
             <% end %>
+            <%= af.file_field :attachment %>
           <% end %>
           </div>
         </div>
--- a/app/views/users/_form.html.erb
+++ b/app/views/users/_form.html.erb
@@ -98,7 +98,6 @@
           <div class="controls">
           <%= f.fields_for :agreement do |af| %>
             <% if af.object.persisted? %>
-              <%= af.file_field :attachment %>
                 <span>Foster Agreement on File, overwrite?</span>
             <% end %>
             <%= af.file_field :attachment %>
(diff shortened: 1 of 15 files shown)


markottaviani:
Keep getting this error:
screenshot_2016-03-26-16-10-53 833

markottaviani:
We already have an "agreement" with is a polymorphic "has_one" of an Attachment class.  Trying to add a second but call it a "confidentiality_agreement", also a "has_one" of an Attachment class.

TheDudeWithTheThing:
Are you initializing that anywhere? On update it might be ok when one exists but on a brand new user it will have none and af.object is nil. :bomb:

markottaviani:
Jackpot!

markottaviani:
Actually.. maybe not hummm

markottaviani:
Okay hitting this issue now, so I'll need to apply the fix mentioned here next.  

http://stackoverflow.com/questions/20966110/rails-4-multiple-polymorphic-has-one-relations-not-working

TheDudeWithTheThing:
multiple `has_one`, one per type you want to support?  That sounds fine, I think you'll need a hidden var to pass the type up in the form too

markottaviani:
Check this out, I think we're done!

app/views/users/_form.html.erb
```diff
@@ -100,7 +100,22 @@
...
+          </div>
+        </div>
+      <% end %>
+
+      <% if is_admin? %>
+        <div class="control-group">
+          <%= f.label :confidentiality_agreement, 'Confidentiality Agreement', :class => 'control-label' %>
+          <div class="controls">
+          <%= f.fields_for :confidentiality_agreement do |af| %>
+            <% if af.object.new_record? %>
```
TheDudeWithTheThing: could clean this up a pinch

```
<% if af.object.persisted? %>
  <span>Confidentiality Agreement on File, overwrite?</span> <%= af.file_field :attachment %>
<% end %>
<%= af.file_field :attachment %>
```

db/migrate/20160412010043_set_agreement_types_for_attachable.rb
```diff
@@ -0,0 +1,9 @@
+class SetAgreementTypesForAttachable < ActiveRecord::Migration
+  def up
+    execute "UPDATE attachments SET agreement_type = 'foster' WHERE attachable_type = 'User';"
```
TheDudeWithTheThing: why the raw SQL?

```ruby
  Attachment.where(attachable_type: 'User').update_all(agreement_type: 'foster')
```

TheDudeWithTheThing:
Good stuff, just some "the more you know" type comments.

:+1:

markottaviani:
Added your suggestions!  Does the ```object.persisted``` stuff look alright?

app/views/users/_form.html.erb
```diff
@@ -97,11 +97,11 @@
           <%= f.label :agreement, 'Foster Agreement', :class => 'control-label' %>
           <div class="controls">
           <%= f.fields_for :agreement do |af| %>
-            <% if af.object.new_record? %>
+            <% if af.object.persisted? %>
               <%= af.file_field :attachment %>
```
TheDudeWithTheThing: you'll have 2 input fields if you leave this no?  you basically want the conditional to show the overwrite? if one exists already right?  trying to remove having to have the line `<%= af.file_field :attachment %>` twice

db/migrate/20160412010043_set_agreement_types_for_attachable.rb
```diff
@@ -1,9 +1,9 @@
 class SetAgreementTypesForAttachable < ActiveRecord::Migration
   def up
-    execute "UPDATE attachments SET agreement_type = 'foster' WHERE attachable_type = 'User';"
+    Attachment.where(attachable_type: 'User').update_all(agreement_type: 'foster')
   end
 
   def down
-    raise ActiveRecord::IrreversibleMigration
+    Attachment.where(attachable_type: 'User').update_all(agreement_type: nil)
```
TheDudeWithTheThing: nice `down` !

app/views/users/_form.html.erb
markottaviani: whoops, yeah.  updated!

markottaviani:
Okay, I think we're golden

Issue #434, opened 2016-01-30
Use ActiveJob for Mailchimp Actions


markottaviani:
http://aspiringwebdev.com/mailchimp-and-active-job-on-rails-4-adding-users-to-your-mailing-list-in-the-background/

PR #449, opened 2016-04-16
WIP #434 ActiveJob Mailchimp attempt

markottaviani:
So, this doesn't seem to work.  I'm getting the following error in delayed_job

````
undefined local variable or method `new' for #<UserSubscribeJob:0x007fd1bb4ec4f8>
/Users/mark/rails_projects/RescueRails/app/jobs/user_subscribe_job.rb:8:in `perform'
/Users/mark/.rvm/gems/ruby-2.2.3/gems/activejob-4.2.6/lib/active_job/execution.rb:32:in `block in perform_now'
/Users/mark/.rvm/gems/ruby-2.2.3/gems/activesupport-4.2.6/lib/active_support/callbacks.rb:117:in `call'
````

I'm not sure where the ```new``` method you use in ```MailChimpService``` is actually coming from

I'm also guessing there is a much cleaner way to handle this, but one step at a time.

### DIFF
--- a/app/jobs/user_subscribe_job.rb
+++ b/app/jobs/user_subscribe_job.rb
@@ -3,6 +3,7 @@ class UserSubscribeJob < ActiveJob::Base
 
   def perform(name, email)
     klass = Rails.env.test? ? FakeMailChimpClient : MailChimpClient
-    klass.user_subscribe(name, email)
+    @client = klass.new
+    @client.user_subscribe(name, email)
   end
 end
--- a/app/jobs/user_subscribe_job.rb
+++ b/app/jobs/user_subscribe_job.rb
@@ -2,8 +2,6 @@ class UserSubscribeJob < ActiveJob::Base
   queue_as :default
 
   def perform(name, email)
-    klass = Rails.env.test? ? FakeMailChimpClient : MailChimpClient
-    @client = klass.new
-    @client.user_subscribe(name, email)
+    MailChimpService.user_subscribe(name, email)
   end
 end
--- a/app/jobs/user_subscribe_job.rb
+++ b/app/jobs/user_subscribe_job.rb
@@ -1,0 +1,10 @@
+class UserSubscribeJob < ActiveJob::Base
+  attr_reader :client
+  queue_as :default
+
+  def perform(name, email)
+    klass = Rails.env.test? ? FakeMailChimpClient : MailChimpClient
+    @client = klass.new
+    new.client.user_subscribe(name, email)
+  end
+end
--- a/app/jobs/user_subscribe_job.rb
+++ b/app/jobs/user_subscribe_job.rb
@@ -1,10 +1,8 @@
 class UserSubscribeJob < ActiveJob::Base
-  attr_reader :client
   queue_as :default
 
   def perform(name, email)
     klass = Rails.env.test? ? FakeMailChimpClient : MailChimpClient
-    @client = klass.new
-    new.client.user_subscribe(name, email)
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by markottaviani** [comment]
This variation gets this error:

```
undefined method `user_subscribe' for MailChimpClient:Class
/Users/mark/rails_projects/RescueRails/app/jobs/user_subscribe_job.rb:7:in `perform'
/Users/mark/.rvm/gems/ruby-2.2.3/gems/activejob-4.2.6/lib/active_job/execution.rb:32:in `block in perform_now'
```

**Comment by markottaviani** [comment]
Okay, was able to implement ActiveJob for the User Subscribe job.  

At this point I think the main question is how to keep it clean.  It would appear I need to put

```
    klass = Rails.env.test? ? FakeMailChimpClient : MailChimpClient
    @client = klass.new
```

As the first 2 lines in each Job, but I'm sure there is a better way to handle this.

**Comment by TheDudeWithTheThing** [comment]
Will try to have a look by end of day

File: app/jobs/user_subscribe_job.rb [hunk 1]
```diff
@@ -0,0 +1,9 @@
+class UserSubscribeJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(name, email)
+    klass = Rails.env.test? ? FakeMailChimpClient : MailChimpClient
+    @client = klass.new
```
markottaviani: @TheDudeWithTheThing would it make sense to put:

```
klass = Rails.env.test? ? FakeMailChimpClient : MailChimpClient
@client = klass.new
```

...into an ActiveJob Concerns file, [as described here](http://stackoverflow.com/questions/35305960/dry-in-activejobbase)?  That way we aren't repeating it in the three other Jobs.

TheDudeWithTheThing: Definitely a way to go but we can maybe do better

TheDudeWithTheThing: If mailing is broken now I'd say go the concern route just to get practice with them and we can revisit.

TheDudeWithTheThing:
Oh wait we have a mailchimp service that does all that for you, how come you aren't using that?

app/jobs/user_subscribe_job.rb
```diff
@@ -0,0 +1,9 @@
...
+    @client.user_subscribe(name, email)
```
TheDudeWithTheThing: This whip thing should be 

MailChimpService.user_subscribe(name, email)

TheDudeWithTheThing:
I'm on the train without my laptop, will have a look once I get to work. I'm not sure why the service isn't working as expected and why tests aren't failing to tell us why.

markottaviani:
The mailchimp service code works, the main thing I'm trying to do is add
ActiveJob support so the subscribe actions go into a queue.  I've had
adoption applications error out when the connection to the mailchimp server
times out.

TheDudeWithTheThing:
Ok so you're doing things a bit backwards and that could be the problem.

Wherever you're trying to send mail you should be calling the Job, which calls the service.

It looks like you've changed the MailchimpService to call the job which is, well, not good.

markottaviani:
Okay I should be doing....

User Model calls the Job
Job calls MailChimpService
MailChimpService calls MailChimpClient
MailChimpClient invokes Gibbon and Hits the MailChimp API

markottaviani:
Reworked it based on your suggestions, yeah this looks MUCH better

TheDudeWithTheThing:
NAILED it

PR #451, opened 2016-04-21
Patches security vulnerability

TheDudeWithTheThing:
This patches the security fail reported by Hakiri.  The tests all pass with this version but bumping jquery-rails up to 4.xxx makes tests fail so that makes me believe we're ok here. I'd give a quick manual check just in case though.

Found here:
https://hakiri.io/github/ophrescue/RescueRails/master/54d9c582629e7206899a165a91b4a38d956e4589/warnings?name=Cross-Site+Request+Forgery

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem 'coffee-rails'
 gem 'uglifier'
 
 gem 'chosen-rails', '~> 1.4.3'
-gem 'jquery-rails', '~> 2.0.3'
+gem 'jquery-rails', '~> 3.1.3'
 gem 'jquery-ui-rails', '~> 2.0.2'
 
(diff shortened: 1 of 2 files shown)


TheDudeWithTheThing:
hmm maybe not, looks like some tests failing on travis, NEEDS WORK

markottaviani:
We're using an ancient version of jquery.  I had looked into this a while back and I think at least part of it had to do with the bbq.js in my vendor library and possibly the jquery.form.wizard.js that's there as well.  

There were some methods used in those that were totally deprecated if I recall correctly, though these errors in Travis might be something else completely.

TheDudeWithTheThing:
killing this, let me know if there is anything else I should tackle

Issue #433, opened 2016-01-30
Upgrade AWS SDK


markottaviani:
AWS SDK2.0 is out, paperclip 5.0 will require it, and the 4.x now supports it.  

Getting paperclip to work with AWS SDK seems fairly easy, however I had a terrible time trying to rewrite the download_url method in the Attachment model.  

@TheDudeWithTheThing any interest in taking this on?

TheDudeWithTheThing:
Will have a look!

TheDudeWithTheThing:
I can't even get AWS to be configured, it's not recognizing the region I'm setting, awful

markottaviani:
Gross, I think we  have a test region you can try will be able to check tomorrow.

TheDudeWithTheThing:
K I'm dumb and was editing the wrong file  (attachements.rb vs. photos.rb)

All good now, just need to configure paths to match the old paths then I think everything, including the file download, will work.

PR #455, opened 2016-07-13
Upgrade gem aws sdk

TheDudeWithTheThing:
WIP for #433


TheDudeWithTheThing:
@markottaviani this should be good to go now, be on the lookout for old files still showing up. I think my buckets are set up different than yours so things were harder to test wrt that.

markottaviani:
Niiiiiiiice

PR #456, opened 2016-07-15
Removes dead code

TheDudeWithTheThing:
I noticed this dead stuff while trying to do the endless scroll.  Having the JSON format in the dogs#index was blocking me from doing the endless scroll the easy way.  If we don't need that dog token stuff (which was commented out) then let's lose it.

Some rubocop clean up too.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -21,7 +21,6 @@
 //= require msk_input.cfg
 //= require jquery.tokeninput
 //= require rescue_rails
-//= require dog_token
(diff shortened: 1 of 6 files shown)


markottaviani:
Looks good!

TheDudeWithTheThing:
haha, I wasn't done yet!  a good amount of clean up to do in the dogs controller, new PR this weekend

markottaviani:
Whoops!

How should we handle tagging PRs so I know what's ready and what's not? "WIP -" in the title?

Mark

PR #458, opened 2016-07-16
Adds endless scrolling to dog search

TheDudeWithTheThing:
Both gallery and manager views.

Not as much test coverage as I like but you can at least play with it and see how she rides.

### DIFF
--- a/app/assets/javascripts/rescue_rails.js
+++ b/app/assets/javascripts/rescue_rails.js
@@ -23,3 +23,16 @@ window.RescueRails = {
   }
 };
 
+$(document).ready(function() {
+  if ($('ul.pagination').length) {
+    $(window).scroll(function() {
+      var url = $('ul.pagination .next_page a').attr('href');
+      if ((url && url !== '#') && $(window).scrollTop() > $(document).height() - $(window).height() - 50) {
+        $('ul.pagination').text("Please Wait...");
+        return $.getScript(url);
+      }
+    });
+    return $(window).scroll();
+  }
+});
+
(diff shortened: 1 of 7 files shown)


markottaviani:
Sweet!  First impression on Chrome is triggering to load more stuff should be a bit more eager. 

Currently possible to be looking at was appears to be the last records without hitting the bottom of the scroll bar to trigger the load.    Also should probably load more on the first go.

TheDudeWithTheThing:
A bit more eager meaning when the last results are visible it's already loading?

Also we should probably stick to the 30 per page, I set it to 9 so I didn't have to create 100+ dogs locally to test.

markottaviani:
Correct on the eager loading.

markottaviani:
Looks good!   The pagination header/footer should probably get removed and I might add a little spinny loading gif if I can find something cool.  

I'm thinking I'll want to load this over in Staging for some folks to play with to make sure they like it as well.

TheDudeWithTheThing:
cool, the pagination is currently what's used to get the URL from for the next page, there might be a way to hide it but removing it would mean we recreate some free stuff we're getting from `will_paginate`

TheDudeWithTheThing:
any other feedback here?

markottaviani:
hmm, not really.  I'm going to push this up to staging and let some folks mess around with it, and try to get some feedback.

markottaviani:
one bit of feedback from one user.  I think dogs should be tweaked to return a multiple of three on pagination, otherwise we get these gaps in the blocks when scrolling.

76633cf9-8f44-4243-a1ec-386d1c7870c0

markottaviani:
Might just be showing a little weird from the lack of photos.  Gunna merge this, push it and see how it goes!

PR #459, opened 2016-07-17
Dog controller tests

TheDudeWithTheThing:
Call me Mayor Goldie Wilson cause I'm cleaning up this town

### DIFF
--- a/app/controllers/dogs_controller.rb
+++ b/app/controllers/dogs_controller.rb
@@ -54,10 +54,10 @@ class DogsController < ApplicationController
   autocomplete :breed, :name, full: true
 
   before_filter :authenticate, except: %i(index show)
-  before_filter :edit_dog_check, only: %i(edit update)
-  before_filter :add_dogs_user, only: %i(new create)
   before_filter :admin_user, only: %i(destroy)
+  before_filter :add_dogs_user, only: %i(new create)
   before_filter :load_dog, only: %i(show edit update destroy)
+  before_filter :edit_dog_check, only: %i(edit update)
 
   def index
     is_manager = signed_in? && session[:mgr_view]
--- a/app/controllers/dogs_controller.rb
+++ b/app/controllers/dogs_controller.rb
@@ -49,7 +49,7 @@
 #
 
 class DogsController < ApplicationController
-  helper_method :is_fostering_dog?
+  helper_method :fostering_dog?
 
   autocomplete :breed, :name, full: true
 
@@ -115,11 +115,7 @@ class DogsController < ApplicationController
   end
 
   def switch_view
-    if !session[:mgr_view]
-      session[:mgr_view] = true
-    else
-      session[:mgr_view] = false
-    end
+    session[:mgr_view] = !session[:mgr_view]
 
     redirect_to dogs_path
   end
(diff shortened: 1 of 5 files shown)


app/controllers/dogs_controller.rb
```diff
@@ -117,130 +115,112 @@ def destroy
...
+  end
 
-    def admin_user
-      redirect_to(root_path) unless current_user.admin?
-    end
+  def admin_user
+    redirect_to(root_path) unless current_user.admin?
+  end
 
-    def is_fostering_dog?(*args)
```
TheDudeWithTheThing: this method was doing some weird stuff, take a look and make sure I didn't trim too much but it seems like its only job should be to compare the `@dog.foster_id` to the `current_user.id`. Also I saw no calling of this method with args, making that branch pretty much worthless.

app/controllers/dogs_controller.rb
```diff
@@ -49,15 +49,15 @@
 #
 
 class DogsController < ApplicationController
-  helper_method :is_fostering_dog?
+  helper_method :fostering_dog?
```
TheDudeWithTheThing: rubocop / the ruby community says "stop prefixing stuff with `is_`"

app/controllers/dogs_controller.rb
```diff
@@ -95,11 +95,9 @@ def update
   end
 
   def create
-    @foster_users = User.all
     @dog = Dog.new(dog_params)
-    if @dog.tracking_id.blank?
-      @dog.tracking_id = Dog.connection.select_value("SELECT nextval('tracking_id_seq')")
-    end
+    @dog.tracking_id = Dog.next_tracking_id if @dog.tracking_id.blank?
```
TheDudeWithTheThing: this could be moved to a default value in the DB instead of the application having to worry about it, maybe make a ticket?

app/controllers/dogs_controller.rb
```diff
@@ -117,130 +115,112 @@ def destroy
...
-                                    :status,
-                                    :age,
-                                    :size,
-                                    :is_altered,
-                                    :gender,
-                                    :is_special_needs,
-                                    :no_dogs,
-                                    :no_cats,
-                                    :no_kids,
+  def dog_params
```
TheDudeWithTheThing: I'm not a fan of indenting private methods twice so the following all looks like it was modified but was mostly outdented 2 spaces.

https://github.com/ophrescue/RescueRails/pull/459/files?w=1 to see without whitespace changes

app/controllers/dogs_controller.rb
markottaviani: Okay, not sure what the common practice is for this type of thing.

#461 created

markottaviani:
I'm good with all of this.  Anything else you wanted to add or shall I merge?

TheDudeWithTheThing:
nope this is all the clean up for dogs currently

TheDudeWithTheThing:
oh wait, bug!

TheDudeWithTheThing:
k all good, I'd say create a dog quick on staging to check before deploying

markottaviani:
Created a few dogs in staging, sequence ID is working well. 

👍

PR #464, opened 2016-07-21
fixes #463 s3 api error on event photos

### LINKED ISSUE #463: S3 Api Error Unable to Add/Modify Event Photo
Aws::S3::Errors::InvalidArgument

To replicate, create or modify an event, and try to add a picture to it.

markottaviani:
caught this in the error logs.  Looks like after the sdk upgrade it's supposed to be ```public-read``` now instead of ```public_read```.  From the paperclip s3 docs, the default is public read anyway, so just deleted the line!


TheDudeWithTheThing:
Oops thought I fixed these, good catch!

PR #466, opened 2016-07-26
Adds a height to the li element of gallery

TheDudeWithTheThing:
## BEFORE:

operation_paws_for_homes___our_dogs

## AFTER:

operation_paws_for_homes___our_dogs

### DIFF
--- a/app/assets/stylesheets/custom.scss
+++ b/app/assets/stylesheets/custom.scss
@@ -140,6 +140,10 @@ form .control-group.error span.error {
   padding-top: 22px;
 }
 
+.gallery li {
+  height: 478px;
+}
+
 img.floatLeft {
     float: left;
     margin: 4px;
@@ -241,6 +245,10 @@ div .fb-like {
 
 .mb-5 {
   margin-bottom: 5px;
+}
+
+.thumb-box {
+  min-height: 478px;
 }
 
 .thumbnail-no-border {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/custom.scss
```diff
@@ -140,6 +140,10 @@ form .control-group.error span.error {
   padding-top: 22px;
 }
 
+.gallery li {
+  height: 478px;
```
TheDudeWithTheThing: 478 is the height of a "normal" gallery box

markottaviani:
Looks good in Chrome, but you wind up with big gaps when viewing on an iPhone.

I can set up with our https://www.browserstack.com login if you don't already have it.   

Might need to go back to the bootstrap grid as per the original dog index: https://github.com/ophrescue/RescueRails/blob/93fc8f66376cd3caf4ebcde3771ff59988c7b198/app/views/dogs/index.html.erb#L27

TheDudeWithTheThing:
Yup that's a pain in the ass though since you have to chunk them 3 at a time, I'll see what I can do.

markottaviani:
Ideally this all goes away after we migrate to newer bootstrap/ Unity / site redesign anyway.  So no worries with it not being perfect.

Mark

TheDudeWithTheThing:
@markottaviani k this should be good now

markottaviani:
👍

PR #468, opened 2016-07-28
Adds searching by dog microchip

TheDudeWithTheThing:


markottaviani:
Niiiiiice

Will test this out tonight and deploy

markottaviani:
From looking at some sample microchips they might be an integer.  In those situations, this will throw errors because ```tracking_id_search``` gets triggered.  

A few example microchips: 
```
982001402146472
0A13212243
```

TheDudeWithTheThing:
Yeah I hate that part, we can just have the search term look through all 3 columns, kinda blah but should be ok

TheDudeWithTheThing:
changing now

TheDudeWithTheThing:
k we're good, just added it to the int search too, we're making a mess here!

markottaviani:
hmm, long ints blow this up.

```
PG::NumericValueOutOfRange: ERROR:  value "982000402146472" is out of range for type integer
LINE 1: SELECT  "dogs".* FROM "dogs" WHERE (tracking_id = '982000402...
                                                          ^
: SELECT  "dogs".* FROM "dogs" WHERE (tracking_id = '982000402146472' OR dogs.microchip = '982000402146472')  ORDER BY tracking_id asc LIMIT 30 OFFSET 0
```

TheDudeWithTheThing:
The only other way is to introduce a search field for microchips or truncate the input for the tracking_id field

markottaviani:
Still blowing up, added a test to demo.  Going to muck with it a little bit

markottaviani:
Okay! Now I think we're good to go.

PR #471, opened 2016-07-28
fixes #469 by making infinite scroll work for users views

### LINKED ISSUE #469: User View stuck on Please Wait..
User view seems to be picking up some of the Infinite Scroll JavaScript, but not returning results.

Fix should be to either get the User/Staff Index View working with Infinite Scroll, or adjust it so that the old Pagination Method works correctly.

markottaviani:
Looked at how to implemented this for the dog view and went ahead and added the extra javascript for it to work for users too!

### DIFF
--- a/app/views/users/_user_table.html.erb
+++ b/app/views/users/_user_table.html.erb
@@ -10,7 +10,7 @@
         <th>Coordinated Dogs</th>
       </tr>
     </thead>
-    <tbody>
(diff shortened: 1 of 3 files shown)


markottaviani:
Gunna go ahead and push this out since it fixes the problem, feel free to post deploy comment if there is anything, but this is pretty much your same code word for word.

PR #472, opened 2016-07-29
Adds dogs-index scope to tie infinite scroll to

TheDudeWithTheThing:


TheDudeWithTheThing:
I don't really like this solution but will be a quick fix, we can do a better one after this is out

markottaviani:
Check out #469 ? Not sure if we still need this or not

TheDudeWithTheThing:
all good then, closing!

PR #478, opened 2016-08-04
Fixes manager view for infinite scroll

TheDudeWithTheThing:
@markottaviani k manager and regular view look good now, sorry bout that.


markottaviani:
Thanks!

PR #483, opened 2016-08-06
added gender and age to dog gallery

markottaviani:
also deleted unused _dogs.html.erb file


TheDudeWithTheThing:
good call on removing `_dog.html.erb`

app/views/dogs/_dog_rows.html.erb
```diff
@@ -15,11 +15,12 @@
               <h4><%= dog.name.titleize %></h4>
               <p>
               <strong><%= dog.status.titleize %></strong><br />
-              <%= dog.primary_breed.name.titleize unless !dog.primary_breed %><br />
-              <%= dog.secondary_breed.name.titleize unless !dog.secondary_breed %><br />
               <% if dog.foster && !dog.foster.city.blank? && !dog.foster.state.blank? %>
                 <%= dog.foster.city.titleize %>, <%= dog.foster.state.upcase %><br />
               <% end %>
+              <%= dog.gender unless !dog.gender %>  <%= dog.age.titleize unless !dog.age %><br />
```
TheDudeWithTheThing: why `unless !` instead of just `if dog.gender` ?

TheDudeWithTheThing: same goes for all the other `unless !`s

markottaviani: This is a common issue for a lot of the views, not sure what the best method for handling fields that might be nil, espically once stuff like ```titleize``` comes in to play

A few items on the topic:

http://codereview.stackexchange.com/questions/28610/handling-nil-trying-to-avoid-try
https://github.com/tokland/tokland/wiki/RubyIdioms#user-content-managing-nils-the-maybe-pattern

Seems like ``` dog.primary_breed.name.try(:titleize) ```  might be a nicer way to do these.

For ```dog.gender``` I don't think I need any logic really, as I'm happy to dispay blank and I don't think Ruby will crash.

markottaviani:
Okay, I think this looks a lot better, check it out.

TheDudeWithTheThing:
seems good  👍

PR #484, opened 2016-08-07
sets decending order for adoption apps

markottaviani:


TheDudeWithTheThing:
very strange, thought I added this in the original infinite scroll PR.

Anyway this DRYs up the searcher a bit #485

Issue #486, opened 2016-08-09
Fix Adoption App Sorting


markottaviani:
Get it to work like Dogs Sorting / Filtering

Or Maybe get those Filter Buttons to Work

markottaviani:
PR #491

PR #491, opened 2016-08-10
#486 Adoption Sort

markottaviani:
@TheDudeWithTheThing  I'm Attempting to implement to sort by clicking on header of a column that we have working for the Dogs Manager Index view that uses the ```sortable``` method in our application helper, however I'm having a tricky time trying to figure out how to use it to sort by an association, specifically the "Assigned To" field.  

Thanks!


TheDudeWithTheThing:
working version added, will let you clean up / write tests

Issue #475, opened 2016-08-02
Infinite Scroll should remember location


markottaviani:
If a user scrolls a few pages down and clicks on a link and then hits the back button, the user should be returned to the same location on the scroll.

TheDudeWithTheThing:
Any idea what browser this isn't working on?  Seems like firefox, chrome and safari on a mac all behave as expected

markottaviani:
On Chrome it is isn't working correctly on the users page.

So If I'm on https://ophrescue.org/users

Scroll down a page or two

Then click on a Dog that someone is fostering, and then hit the browsers back button I wind up at the top of the page.

markottaviani:
Also requested a supportdetails.com report from the person that reported the issue.

markottaviani:
User is on Firefox 48 / Windows 10.

I'm seeing the same issues though.  It seems like the scroll bar is going to the right position, but the data for the lower "pages" of info isn't getting re-queried right?

PR #492, opened 2016-08-11
Remember infinite scroll position

TheDudeWithTheThing:
This method uses the HTML History API and replaces the current url with URL?page=# in the user's history. When the user hits back they'll be on the page they last had load from the infinite scroll but the only way to go back pages is to click the page numbers.

This seems like a good interim solution and is generic enough to work for wherever we're doing infinite scroll currently.

Give her a test ride!

### DIFF
--- a/app/assets/javascripts/rescue_rails.js
+++ b/app/assets/javascripts/rescue_rails.js
@@ -19,7 +19,7 @@ window.RescueRails = {
       $status.addClass('fa fa-times');
       $status.show();
       $status.fadeOut(3000);
-    })
+    });
   }
 };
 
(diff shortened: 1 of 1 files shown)


TheDudeWithTheThing:
Fixes #475

markottaviani:
Hmm, doesn't seem to be working quite right.  

I'm going to try and do some research tonight, looks like a common issue:

https://www.google.com/#q=infinite+scroll+back+button+solution

markottaviani:
If these get to be too much of a hassle, it might be worth limiting where we do the infinite scroll stuff.  Possibly remove from the table-based listings and use exclusively in the Dog Gallery.

TheDudeWithTheThing:
Yeah the Internet didn't seem to have a solid solution to this

markottaviani:
Thanks a lot internet!

markottaviani:
So, just read through this: https://www.smashingmagazine.com/2016/03/pagination-infinite-scrolling-load-more-buttons/

I'm leaning more toward switching the tables that the staff uses to keep track of stuff back to pagination, and then for the dog gallery possibly a "Show More Dogs" button.  Maybe adjust the number of results returned based on mobile vs desktop?

TheDudeWithTheThing:
Yeah since we don't have a ton of time to sink into this and I'm no front end expert then I'd be fine with ripping it out, let's kill it

markottaviani:
A++

PR #496, opened 2016-08-15
Adopter sorts ez

markottaviani:
Quick enable of sortable for adoption columns that don't depend on a join.


markottaviani:
Will be merging and pushing once the Ci build wraps up.

Issue #497, opened 2016-08-15
Sort Adopters By Last Comment Fields


markottaviani:
On Adopters Index View, adjust the sortable method to handle the "Last Comment Timestamp" and the "Last Comment" fields

PR #498, opened 2016-08-16
Sort Adopters by Last Comment on each App?

markottaviani:
This is an attempt at #497 However it doesn't work right, and I'm not sure exactly how to handle this sort, as I want sort adopters by the timestamp of the most recent comment on their application.

### DIFF
--- a/app/models/adopter_searcher.rb
+++ b/app/models/adopter_searcher.rb
@@ -38,8 +38,11 @@ class AdopterSearcher
 
   def with_sorting
     if @params[:sort] == 'assigned_to'
-      with_join_for_sort
+      with_join_users_for_sort
       column = 'users.name'
+    elsif @params[:sort] == 'comments.updated_at'
+      with_join_comments_for_sort
+      column = 'comments.updated_at'
     else
       column = sort_column
     end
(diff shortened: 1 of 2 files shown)


TheDudeWithTheThing:
@markottaviani close!

https://github.com/ophrescue/RescueRails/pull/498/commits/dcdaadc9bf8dc39b00e82a54d91207fdb8c80cec

markottaviani:
hmm, still doesn't seem to be working right.

TheDudeWithTheThing:
oh yeah, we need to select out the comment with the max(updated_at) and then sort by that, could be tricky

http://stackoverflow.com/a/17179241

MatthewDrouin:
IDK what the database structure is (if there is an easy way to get it let me know) but you should be able to join on a sub select and sort by the updated_at.

SELECT app.*, c.updated_at 
FROM app 
LEFT JOIN (SELECT * FROM comments WHERE 1=1 GROUP BY id ORDER BY comments.updated_at) AS c ON app.id = c.app_id
ORDER BY c.updated_at

Do a left join because I assume you want apps even if there are no comments.

MatthewDrouin:
GROUP BY id should be GROUP BY app_id

TheDudeWithTheThing:
welcome @MatthewDrouin !  from out of nowhere

it's not pretty but it's doing the job, matt was on the right track, let me make sure we're getting good index usage, there might be a more railsy way to do it too

MatthewDrouin:
The rails way of doing it is to probably add a last_comment_date to the app and update that but I never like those solutions :)

markottaviani:
Sweeet @MatthewDrouin for the win!  This is working pretty well.   Going to go ahead and merge.

MatthewDrouin:
Glad I could be of assistance :)

PR #503, opened 2016-08-28
Fixes mailto link icon

TheDudeWithTheThing:
During the rails 5 upgrade a test kept failing because icon_mail.gif was
missing.  I'm not sure why rails 4 let this go without reporting a missing image but rails 5 is VERY ANGRY.

Currently broken on production, check out the contact page, https://ophrescue.org/contact

The after


markottaviani:
haha, well let's make Rails Happy.

PR #506, opened 2016-08-29
Removes unused json_spec gem

TheDudeWithTheThing:
previously in the rails-5 branch, let's dump this thing on its own.


TheDudeWithTheThing:
this is already part of removing asset gems, I'm starting to confuse myself

PR #508, opened 2016-08-30
Adds jquery_ujs to fix CSRF bug

TheDudeWithTheThing:
From bug reported.  Ajax requests were missing the X-CSRF-Token header from removing jquery-ujs


markottaviani:
Thanks! 

Mark

PR #509, opened 2016-08-31
Fix responsive bootstrap

TheDudeWithTheThing:
Ripped the assets out of the gem we were using for this.  Tested on my phone locally and all looks good.


TheDudeWithTheThing:
@markottaviani this is good to go

markottaviani:
tested as well, looks good. Thanks!  Merging and Deploying!

Issue #524, opened 2016-11-27
Comment Post button won't re-enable


markottaviani:
When I update an application, by posting a comment - I can't post another comment until I refresh the screen. The post button will not work.

PR #525, opened 2016-11-27
fix for #524 renable commit button after post

markottaviani:
Was able to verify that for comments on dogs or adopters after a comment was posted, the Post button wouldn't re-enable and folks have to refresh the page for the post button to start working again.

This appears to fix the issue.  Though I'm not sure why it started in the first place, or if there is a cleaner way to handle it.   

Thanks!

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -36,6 +36,9 @@ $( function () {
 
         // clear comment field for next comment
         $('#comment_content').val('');
+
+        // Turn the POST button back on
+        $('#comment_submit').prop('disabled', false);
       },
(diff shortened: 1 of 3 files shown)


TheDudeWithTheThing:
hmm weird, I think the page used to refresh maybe? Will give a closer look tomorrow

TheDudeWithTheThing:
What's really strange is there's nothing obvious that is disabling the button

TheDudeWithTheThing:
I think we're ok to push this just to unblock / unannoy people, maybe make an issue to revisit?

markottaviani:
Sounds good!  Thanks!

PR #526, opened 2016-11-29
fix for link dog button getting stuck as disabled

markottaviani:
Fairly straight forward, not sure if specifying an ID for the button is "best practice" or I should get at the buttom via DOM traversal a different way.


TheDudeWithTheThing:
It's fine if it's in fact unique.

I wonder what the change is in fails 5 that is disabling everything

TheDudeWithTheThing:
Also all good 

👍

PR #532, opened 2016-12-18
closed #531 reference relationship checker

### LINKED ISSUE #531: throw jq validate error if reference is related
>Is there a way to create an error if someone lists a reference as a family member? For example, if they list mother, father, brother, sister - the most common ones? We still have many applications that we need to ask them for replacement references. I know it tells them non family, but people are obviously not reading it.

markottaviani:
Attempt at #531 , some weird behavior where clicking on and off the relationship box causes it to go true, but it at least gets the point across to folks applying.


markottaviani:
@TheDudeWithTheThing *nudge* I totally forgot about this PR, if ya get a chance to take a look that'd be great!

markottaviani:
Upgraded JQuery and tweaked the inline help as well.  Same odd issue with clicking on and clicking off of the field, but it's still an improvement from the current state of affairs and anyone that tries to put their mom down as a reference is still going to get the message that they shouldn't do it.

Will merge and deploy shortly, thanks for taking a look!

PR #545, opened 2017-01-10
closes #540 home verification auth to adopt app

### LINKED ISSUE #540: Add Home Verification Auth
> Would it be possible to add the pop up tab that I mentioned before the holidays? It should pop up when an applicant marks that they own their home. It should say, "I give Operation Paws for Homes permission to verify my homeownership". This should be a hard stop if they do not agree but they should have the option to go back and check rental if needed.

> Laurie L. & Jen D.

markottaviani:
includes show/hide of field if home type of own is set
and display of authorization if required approval was captured.

### DIFF
--- a/app/assets/javascripts/adoption_app_home_auth.js
+++ b/app/assets/javascripts/adoption_app_home_auth.js
@@ -1,11 +1,11 @@
-$(document).ready(function() {
-    $( ".own-home-auth" ).hide();
-    $(".residence-auth-pop").change(function() {
-        if (this.value == 'own'){
-            $( ".own-home-auth" ).show();
-    }
-    else if (this.value == 'rent') {
-        $( ".own-home-auth" ).hide();
-    }
+$(document).ready(function () {
+    $(".own-home-auth").hide();
+    $(".residence-auth-pop").change(function () {
+        if (this.value == 'own') {
+            $(".own-home-auth").show();
+        }
+        else if (this.value == 'rent') {
+            $(".own-home-auth").hide();
+        }
     });
 });
--- a/app/assets/javascripts/adoption_app_home_auth.js
+++ b/app/assets/javascripts/adoption_app_home_auth.js
@@ -4,7 +4,7 @@ $(document).ready(function () {
         if (this.value == 'own') {
             $(".own-home-auth").show();
         }
-        else if (this.value == 'rent') {
+        else {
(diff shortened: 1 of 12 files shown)


markottaviani:
Sanity check on my Jquery should be the main thing with this.

app/assets/javascripts/adoption_app_home_auth.js
```diff
@@ -0,0 +1,11 @@
...
+    $( ".own-home-auth" ).hide();
+    $(".residence-auth-pop").change(function() {
+        if (this.value == 'own'){
+            $( ".own-home-auth" ).show();
+    }
+    else if (this.value == 'rent') {
+        $( ".own-home-auth" ).hide();
+    }
+    });
+});
```
TheDudeWithTheThing: your spacing here is all wacky and no newline at the end

app/assets/javascripts/adoption_app_home_auth.js
```diff
@@ -0,0 +1,11 @@
+$(document).ready(function() {
+    $( ".own-home-auth" ).hide();
```
TheDudeWithTheThing: why the spaces in the parens?

markottaviani:
Thanks, been switching over to VS Code, and hadn't put all the config in yet.

TheDudeWithTheThing:
great tests!  when do we switch to react or one of the other 1000 JS frameworks?

app/assets/javascripts/adoption_app_home_auth.js
```diff
@@ -0,0 +1,11 @@
+$(document).ready(function () {
+    $(".own-home-auth").hide();
+    $(".residence-auth-pop").change(function () {
+        if (this.value == 'own') {
+            $(".own-home-auth").show();
+        }
+        else if (this.value == 'rent') {
+            $(".own-home-auth").hide();
+        }
```
TheDudeWithTheThing: should this just be an else hide?  otherwise it's going to keep whatever state it's in unless the value is own or rent

markottaviani: Well, technically it's a radio button so the only two options "should" be a rent or own.  But logically an else hide is more accurate since we'd always want the field hidden if the value was other than own.  I'll toss the update in.

markottaviani:
Woo Shipped!

Issue #514, opened 2016-11-02
Adopt App Inputs review


markottaviani:
Error on Landlord name over 100 char.  

Add Inline Character Limit tests to LandLord name and all other fields

Also update Landlord name to "LandLord / Property" Name

PR #552, opened 2017-02-22
Switches to jquery UI autocomplete for dog link

### LINKED ISSUE #551: Improve Dog Selection on Mobile for Linking to Applications
> Any way to add a search bar for dog names when linking a dog to an application via mobile browser? I just watched my Mom scroll through allllllllllll the names in DMS to link a dog!

TheDudeWithTheThing:
Fixes #551

dog-autocomplete


markottaviani:
Sweet, great implementation.

And wow we actually had test coverage for the old dropdown!

Some of these test fails about missing a PDF file are a my bad.  From all the brittle UI tests I had ignored some of the Travis CI Runs.

TheDudeWithTheThing:
Yup test fixes coming tonight! Want to make a helper for the new style.

app/controllers/dogs_controller.rb
```diff
@@ -60,10 +60,15 @@ class DogsController < ApplicationController
   before_action :edit_dog_check, only: %i(edit update)
 
   def index
-    is_manager = signed_in? && session[:mgr_view]
+    is_manager = signed_in?
```
TheDudeWithTheThing: this is a little strange, search functionality changes based on mgr_view, should be ok to remove yeah?

TheDudeWithTheThing:
@markottaviani a couple things came out of this PR

1. We're missing a lot of validations, I was able to save an adopter record without a `dog_id`. Going to do a full validation sweep, at least at the app level for now, we can do DB level constraints as a next pass.

2. We have a lot of endpoints doing double duty with `respond_to` formats, views, html, nothing is really consistent. I'm not sure what the solution here is for that yet but will be keeping an eye out for a better way to deal with this. It feels weird to hit /dogs to load html and then the autocomplete also hits that endpoint for the dog list.

app/controllers/dogs_controller.rb
markottaviani: hmm, should be okay.  However the title will probably always display as 'Dog Manager' when someone is signed in.  Probably overloading the use of ```is_manager``` 

Maybe something like

 ```@title = session[:mgr_view] ? 'Dog Manager' : 'Our Dogs' ```

In addition to your change would do it?

markottaviani:
Sounds good thanks, I know there are validation gaps on the adoption application, have a few noted on  #514

TheDudeWithTheThing:
@markottaviani just pushed a change which fixes a pretty huge bug, the error messaging is wrong for sure but probably best to plug this sooner rather than later.

TheDudeWithTheThing:
@markottaviani k I'm all set here, made some other changes since I was in there.

The bug this fixes is right now you can "blank out" a previous dog link by just mashing the Link Dog button. Now that we're doing `find` on both we verify they exist first, in this case it will block any empty `dog_id` coming in.

The error handling here is still weird but should probably just do a full app sweep on that next.

markottaviani:
Okay great, I'm going to see if I can patch up those other failing tests and we can get this merged in.

TheDudeWithTheThing:
are there other chosen spots we should swap this to sooner rather than later?

markottaviani:
I don't think so, but once this change is rolled out we can query the volunteers and see if anywhere else would benefit from it.

markottaviani:
Looks good to me, will ship this tonight.

PR #554, opened 2017-03-02
bugfix if logged in only public view dogs should be in gallery view

### LINKED ISSUE #553: Public Dog View showing every dog ever
(Rolled production back to the prior release so we don't have to rush a patch.  )

Public View of Dogs was showing all dogs regardless of adoption status.  

Todo
* Patch Code
* Add Test Scenarios.

markottaviani:
This should fix #553 which was introduced in PR #552 

Gallery view is supposed to work the same regardless of if you're logged in or not.  We introduced an issue where if logged in all dogs ever would show up in the gallery view.  Totally missed that in my review.  

I've added a whole bunch of tests that should verify behavior.

### DIFF
--- a/app/controllers/dogs_controller.rb
+++ b/app/controllers/dogs_controller.rb
@@ -62,7 +62,9 @@ class DogsController < ApplicationController
   def index
     @title = session[:mgr_view] ? 'Dog Manager' : 'Our Dogs'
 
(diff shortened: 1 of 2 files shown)


TheDudeWithTheThing:
doh

👍

PR #557, opened 2017-03-04
Fix for linking dogs to applications

### LINKED ISSUE #556: Having Mgr View Session Attribute set to false breaks Dog Linker
>Folks, if you're having trouble linking dogs to applications, switch over to Manager view in /dogs and go back into the Adoption Application.

markottaviani:
This fixes #556 a defect where if someone had mgr_view set to false the dog link autocomplete would return limited or weird results


TheDudeWithTheThing:
ah the rubyzip security thing, gotta fix that one up

we really need to remove "permission" logic from the `DogSearcher` I think

👍

app/controllers/dogs_controller.rb
```diff
@@ -62,7 +62,14 @@ class DogsController < ApplicationController
   def index
     @title = session[:mgr_view] ? 'Dog Manager' : 'Our Dogs'
 
-    do_manager_view = signed_in? && session[:mgr_view]
+    if signed_in? && session[:mgr_view]
```
TheDudeWithTheThing: could get slick and do

`do_manager_view = signed_in? && (session[:mgr_view] || params[:all_dogs] == "true")`

spec/controllers/dogs_controller_spec.rb
```diff
@@ -80,6 +80,13 @@
         expect(assigns(:dogs)).to match_array([adoptable_dog, adoption_pending_dog, coming_soon_dog])
       end
 
+      it 'with all dogs paramater set all dogs are returned' do
+        get :index, params: {all_dogs: true}, session: { mgr_view: true }
```
TheDudeWithTheThing: should make mgr_view false so it hits the all dogs branch

markottaviani: Covering that at https://github.com/ophrescue/RescueRails/blob/2ebb0a19684708fe42be30247f369610ae65b1b7/spec/controllers/dogs_controller_spec.rb#L78 I think?

TheDudeWithTheThing:
coveralls says it's covered so fire away!

PR #566, opened 2017-03-06
Adds test for massive flub with adopter edit

TheDudeWithTheThing:
never make that mistake again!   Tests!

Also I think the newer version of capybara makes it so we can remove the sleeps


TheDudeWithTheThing:
haha, no credit needed as that would mean I accept blame for the huge mess  😃

PR #567, opened 2017-03-07
Test Tidy Up and Minor Bug Fix

markottaviani:
* 	needed to add one sleep back for local testing, otherwise got weird stuff like this in the Landlord Phone Number:

* Also fixed 500 error when going to events/:id/:edit and not already being logged in.

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -27,6 +27,7 @@
 #
 
 class EventsController < ApplicationController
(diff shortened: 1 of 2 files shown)


TheDudeWithTheThing:
maybe throw them all back in if they're still flakey?  I ran the tests a bunch locally and passed every time

markottaviani:
Nah, I ran the whole set a bunch of times with and without this specific sleep, always red without it always green with it.    ¯\_(ツ)_/¯

PR #568, opened 2017-03-07
Track dog status

### LINKED ISSUE #562: Track Status Changes to Dogs
Similar to how we have "<user> changed status of application from new to workup" in the Adoption Application comments.  Do the same thing for Dogs.

When someone changes the status of a dog, an entry should be made in the comments section.

TheDudeWithTheThing:
Extracts audit code to an Auditable concern, which made auditing dog status a snap.

Closes #562


markottaviani:
🥇

Issue #576, opened 2017-05-05
Eager Load Dog Public Index


markottaviani:
Dog Public Index is probably our most popular page, but it makes way too many database calls.  

Let's make it faster!

PR #577, opened 2017-05-05
eagar load dog index

markottaviani:
Implements #576

### DIFF
--- a/app/models/dog_searcher.rb
+++ b/app/models/dog_searcher.rb
@@ -17,28 +17,30 @@ class DogSearcher
   end
 
   def search
+    @dogs = Dog.includes(:photos, :foster)
     if @manager
+      @dogs = @dogs.includes(:adopters, :comments)
       if tracking_id_search?
-        @dogs = Dog.where(
+        @dogs = @dogs.where(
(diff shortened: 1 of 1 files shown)


app/models/dog_searcher.rb
```diff
@@ -38,7 +38,7 @@ def search
         @dogs = Dog.where("dogs.name ILIKE ?", "%#{@params[:q]}%")
       end
     else
-      @dogs = Dog.where("status IN (?)", PUBLIC_STATUSES)
+      @dogs = Dog.includes(:photos, :foster, :primary_breed, :secondary_breed).where("status IN (?)", PUBLIC_STATUSES)
```
TheDudeWithTheThing: is this the only spot you need this?  might make sense to change all paths here

you always want @dogs so maybe at line 21 you do

```
@dogs = Dog.includes(:photos, :foster, :primary_breed, :secondary_breed)
```

and then where `@dogs` is now you do `@dogs = @dogs.where(whatever the clause is now)`

markottaviani:
@TheDudeWithTheThing awesome idea and much better!

PR #578, opened 2017-05-13
add medical review flag

markottaviani:


db/migrate/20170513172145_add_medical_review_to_dogs.rb
```diff
@@ -0,0 +1,5 @@
+class AddMedicalReviewToDogs < ActiveRecord::Migration[5.0]
+  def change
+    add_column :dogs, :medical_review_complete, :boolean, default: false
```
TheDudeWithTheThing: for "no down time deployments" this should be moved out of this PR and deployed separately, else there's some time in there where you're trying to access an attribute that doesn't exist.

markottaviani: I think the cap deploy is setup to handle the db updates before the code goes out, so I think we're good.  Thanks!

Issue #584, opened 2017-06-17
Restrict Name and Email Changes


markottaviani:
Require Admin Permissions to edit the name or email address associated with a staff account

Name or Email Address on a user account should not be editable by anyone, including the owner of the User account, unless someone has admin rights.  

Textboxes for Name & email should get the Disabled attribute added to them as per http://getbootstrap.com/2.3.2/base-css.html#forms

```user_params``` on the User Controller should also be updated.  

Notify Mindy & Laurie on Change

**Comment by lnzd17** [comment]
I can take a shot at this one starting Monday morning. Do we want to avoid adding form html/erb directly to the User#edit page? I see that it is being rendered from a template which does not have the Name/Email address fields  set to disabled.

**Comment by markottaviani** [comment]
Sounds good!    Yup, I think editing ```/app/views/users/_form.html.erb``` can cover the UI changes.  ```_form``` gets used for creating new users (which only admins can do anyway) and when admins or the user themselves update their profile.   

So probably for ```_form.html.erb```  you can just leverage ```is_admin?``` : 

```  if is_admin?
enable the fields
 else
disable the fields
```

lnzd17:
I have created a branch field_restrictions_#584 for this ticket. I have implemented the conditional statement in the `_form.html.erb`  file and updated the `user_params` for a non `admin` user. 

I was thinking that this change only needed some type of view test (to check if a field is disabled or not depending on who the user is). I will look into how to do this as I have never written a view test before.

Do I link the *Issue* to my *PR*  by mentioning the ticket number in the *PR* comments? Wasn't sure what the best practice is there.

markottaviani:
> Do I link the Issue to my PR by mentioning the ticket number in the PR comments.

That's what we usually do, thanks!

> I was thinking that this change only needed some type of view test (to check if a field is disabled or not depending on who the user is)  

I agree, we have a good number of view layer tests in the repo already that should serve as a decent baseline.

lnzd17:
PR up :)

PR #598, opened 2017-07-17
Field restrictions #584

lnzd17:
PR for issue #584 
* updated `_form.html.erb` to display disabled name/email fields to non-admin users
* updated `user_params` accordingly
* created feature tests

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -413,7 +413,7 @@ DEPENDENCIES
   will_paginate
 
 RUBY VERSION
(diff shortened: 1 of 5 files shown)


markottaviani:
👍   Looks good to me! 

I'm going to add a inline help for the users.  Otherwise I'll get a bunch of emails from people confused about why they can't change their email address anymore 😄

markottaviani:
Released to prod.

Issue #592, opened 2017-07-15
Sorting after a search removes search criteria


markottaviani:
from the applications or dog manager view pages, perform a search (optional: that yields multiple results), then sort by a column. The search goes away and the columns are sorted without the search criteria.

markottaviani:
Bug Confirmed

PR #600, opened 2017-07-18
added search_term variable  to sortable method

lnzd17:
PR for issue #592 
*added `search_term` to the `link_to` UrlHelper inside the `sortable` method


markottaviani:
connects to #592

PR #602, opened 2017-07-19
Caches dogs at the row level

TheDudeWithTheThing:
Caches at the dog level, not sure how much this will buy us in terms of performance gains but should be ok.  My environment is a little messed up so I can't try this out until I fix that.

### DIFF
--- a/app/models/photo.rb
+++ b/app/models/photo.rb
@@ -30,7 +30,7 @@
 
 class Photo < ApplicationRecord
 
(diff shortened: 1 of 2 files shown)


markottaviani:
Quick fix of adding ```touch``` to Photos.  Gallery cache will now expire correctly when someone adds photos or changes their order for a dog.

TheDudeWithTheThing:
Good call

Issue #583, opened 2017-06-17
Search Staff by Email Address


markottaviani:
via Lisa Taylor email

Edit the search on /users so that it can search the email field or the name field

lnzd17:
@markottaviani  I have a working solution for this ticket - but this solution adds a third search box to the Users#index. Is that acceptable? Or would you prefer we pursue a more condensed solution?

markottaviani:
Cool! I think this can be condensed down with the name search, with a Regular Expression (regex)  

Should be a lot of examples on Stack overflow for an email regex.  

Basically use a Regex to see if there was a valid email address entered into the box (text followed by an @ followed by more text a . and then more text after that) 

 If so, search the email column, if not search the name column.

Should provide a more condensed solution as well as give you a chance to play with Regex if you haven't yet.

lnzd17:
Awesome. I will try that out!

markottaviani:
Cool have fun! 

Also a worth mentioning:

Broken / Half Baked Pull Requests are totally fine.   We usually put "WIP" for Work in Progress at the front of the title when it's still something we're working on or stuck on.

lnzd17:
Makes sense! Will follow that standard going forward.

Issue #192, opened 2013-01-16
Full Text Searching across all data


markottaviani:
Ability to search all DMS Content a user has access to.  Fully Indexed Global Search, dogs, comments, etc.  

Existing search is pretty rough and low-tech.  

Assess Sunspot and other ruby search solutions for improving searching through OPH records.

http://railscasts.com/episodes/278-search-with-sunspot

docvego:
Is there anyway to or could it be added to the task list to search dogs by their micro chip numbers?

We had a call today about a found dog that was one of ours and it would have taken me forever to find him if I hadn't posted on Facebook and the foster knew who it was.

TheDudeWithTheThing:
do we have that data available?

docvego:
Yeah, dog.microchip

TheDudeWithTheThing:
oh man this thing is over a year old

TheDudeWithTheThing:
do we make the commitment to elastic search?  that's more cost to OPH

markottaviani:
No commitment to anything we can use whatever we want.

TheDudeWithTheThing:
sorry commitment is the wrong word, do we make the jump? Time to take `DogSearcher` for his last walk?

TheDudeWithTheThing:
I can bang this out in 30min vs 1 week for moving search to something else

markottaviani:
Certainly a fan of getting this done nice and easy, what would we be dealing with $$$ wise for a move to Elastic?

TheDudeWithTheThing:
cost!
https://aws.amazon.com/elasticsearch-service/pricing/

markottaviani:
@TheDudeWithTheThing I'll talk to the OPH Folks about cost next week on this to get a solid gauge on interest.  

t2.micro.elasticsearch    - staging
t2.small.elasticsearch    - production

Honestly even two micros might be enough.      ~$40 /month on the high side

TheDudeWithTheThing:
What's the usage of the current search?

markottaviani:
Honestly not much.  Only users of the search are OPH Staff/Volunteers.

markottaviani:
Also a possible option https://www.viget.com/articles/implementing-full-text-search-in-rails-with-postgres

markottaviani:
not going to do this one

PR #603, opened 2017-07-20
#583_User_Email_Search

lnzd17:
PR for ticket #583 
* name and email can be searched from the same input field
* created controller test

One thing to note: When searching by email a full email address has to be submitted. It does not work with a 'partial' address. Did we want to have that work? 

I was trying to figure out how to 'clean' up the `params` if statement inside the `index` method. Would love to hear suggestions on how to make it a bit tidier if anything comes to mind.

### DIFF
--- a/app/models/user_searcher.rb
+++ b/app/models/user_searcher.rb
@@ -7,7 +7,7 @@ class UserSearcher
   end
 
   def search
-    @users = User.all
+    @users = User
 
     if text_search?
       if email_search?
--- a/app/models/user_searcher.rb
+++ b/app/models/user_searcher.rb
@@ -1,0 +1,66 @@
+class UserSearcher
+  PER_PAGE = 30
+  EMAIL_CHECK = /\A([\w+\-].?)+@[a-z\d\-]+(\.[a-z]+)*\.[a-z]+\z/i
+
+  def initialize(params: {})
+    @params = params
+  end
+
+  def search
+    @users = User.all
+
+    if text_search?
+      if email_search?
+        @users = @users.where('lower(email) LIKE ?', "%#{search_term}%")
+      else
+        @users = @users.where('lower(name) LIKE ?', "%#{search_term}%")
+      end
+    elsif location_search?
+      @users = @users.active.near(@params[:location], 30)
+    else
+      @users = @users.active.filter(filtering_params).order('name')
+    end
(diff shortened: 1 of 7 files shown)


markottaviani:
Great work thanks!   @TheDudeWithTheThing any ideas on how you might tidy up the users_controller?  I'm not sure either 😄

TheDudeWithTheThing:
come on now, all that stuff in the controller should be moved to a UserSearchService object, the end result would look glorious

```
@users = UserSearchService.search(params)
```

TheDudeWithTheThing:
@lnzd17 there's already a DogSearch class you can "mimic"

markottaviani:
haha, of course.  @lnzd17 I'm sure that'd be a good experience for you if you feel like taking a shot at it.  

We also have #192 looming somewhere in the future, which would make a lot of work here moot.

lnzd17:
I'll see if I can `hack it`. :)

lnzd17:
Created UserSearch class. I will work on tests now. I think this is on the right track. Holler if not :)

app/models/user_searcher.rb
```diff
@@ -0,0 +1,66 @@
+class UserSearcher
+  PER_PAGE = 30
+  EMAIL_CHECK = /\A([\w+\-].?)+@[a-z\d\-]+(\.[a-z]+)*\.[a-z]+\z/i
+
+  def initialize(params: {})
+    @params = params
+  end
+
+  def search
+    @users = User.all
```
TheDudeWithTheThing: no need for the `all` part here

lnzd17: 👍

TheDudeWithTheThing:
this looks great so far, nice work!

lnzd17:
ready for review

app/views/users/_user.html.erb
```diff
@@ -1,5 +1,6 @@
 <tr>
   <td><%= link_to user.name, user %></td>
+  <td><%= user.email %></td>
```
markottaviani: Given that an email search is only going to return one User, and the limited real-estate on the table view,  I'm thinking that we don't display the email address in the table view.

lnzd17: That makes sense. I can revise.

PR #604, opened 2017-07-22
Ready for Review - #594 Multi select dog filter

### LINKED ISSUE #594: Dog Filters for Manager view
Related to #357

Provide Advanced Filtering Options on the Dog Manager Screen

Should Provide
* Quick Identification of Dogs that need Foster Homes
* Lots of other stuff

markottaviani:
closes #594 

- [x] Age Filter should be able to filter by a single age value
- [x] Age Filter should be able to filter by multiple age values
- [x] Size Filter should be able to filter by a single size value
- [x] Size Filter should be able to filter by multiple size values
- [x] Status Filter should be able to filter by a single status
- [x] "Reset to a  Default filter view"  Button
- [x]  Dog filter widget shouldn't reset to **Any** after clicking filter button
- [x] Filtering with multiple filters (e.x. ```age=young and size=small```) at once shouldn't throw an error.  
- [x] Flags Filters Should Work
- [x] Tests should pass
- [x] Additional tests should be written to cover all of the above items


markottaviani:
Currently stuck when trying to filter by two values 

```/dogs?utf8=✓&age=baby&size=small&commit=Filter```

Results in ```wrong number of arguments (given 1, expected 0)``` in ```Filterable```.

TheDudeWithTheThing:
@markottaviani oh easy, you have the param you're filtering called size but the scope is dsize.

Also in the scope you have size as the param name and dsize as the var you pass in the where clause

markottaviani:
Is there some way to modify the ```public_send``` in [filterable ](https://github.com/ophrescue/RescueRails/blob/6fca0dce353b84cee6aae5e633ded1a17dc44a2d/app/models/concerns/filterable.rb#L23)  to be explicit as to the class and method it's trying to call and not just the method?

When I had the scope just named ```size``` it was hitting a size method in ActiveRecord itself.

TheDudeWithTheThing:
It is calling the right method, though I wonder why your scope didn't override the ActiveRecord size. We don't want to do that anyway.

 I wonder if the scopes should be renamed too since they clash with the attribute names.

markottaviani:
Renamed scopes and tidied up a little bit.  

So, next thing I'm not sure about (and admittedly haven't looked at yet) : how to deal with "OR" for a specific filter.   If I wanted ```?is_age=baby&is_age=young``` to return all dogs in either age group.

markottaviani:
Whoops, just needed ```multiple: true``` instead of ```multiple: multiple```

markottaviani:
Okay, I'm really not sure how to approach the Flags Filters.  I can do these ugly looking separate checkboxes as in 88e4eb82ff6e6ec1e8144fb9e81dd2d9570faf2e and everything is easy to scope.    


Options I think are: 
1. A UI challenge of how do I make a bunch of check_box_tags appear like a dropdown multi-select?   Using http://davidstutz.github.io/bootstrap-multiselect/ for the other ones on this filter page

2. Figure out a good way to parse an array of values into separate scopes  

3. Something else I'm not thing of?

markottaviani:
Ping @TheDudeWithTheThing  whenever you get a chance

markottaviani:
Flag Dropdown now works, though the styling is ugly and it closes after every click.  It will need a little bit of polish still.

markottaviani:
Well, from mobile testing the manager view is pretty ugly anyway, so I'm going to pass on the mobile improvements for now.  It's at least functional 😄      Done and ready to merge!

Issue #570, opened 2017-04-06
[ERROR] dogs#index PG invalid input syntax for integer


markottaviani:
Low priority fix

```
An ActionView::Template::Error occurred in dogs#index:

PG::InvalidTextRepresentation: ERROR:  invalid input syntax for integer: "4 seasons"
LINE 1: SELECT COUNT(*) FROM “dogs” WHERE (tracking_id = '4 seasons'…

^
: SELECT COUNT(*) FROM “dogs” WHERE (tracking_id = ‘4 seasons’ OR dogs.microchip = ‘4 seasons’)

app/views/dogs/index.html.erb:20:in `_app_views_dogs_index_html_erb'
------------------------------- Request:
* URL        : https://ophrescue.org/dogs?utf8=%E2%9C%93&search=4+seasons
* HTTP Method: GET
* Parameters : {"utf8"=>"✓", "search"=>"4 seasons", "controller"=>"dogs", "action"=>"index"}
```

lnzd17:
Hi Mark - does this error only occur when a user searches for a dog's name that includes both numbers and letters?

markottaviani:
Correct 

Specifically if it starts with a number

PR #606, opened 2017-07-27
4 seasons now works

lnzd17:
PR for #570 
* dogs that have a name with a number at the beginning can now be searched

### DIFF
--- a/app/models/dog_searcher.rb
+++ b/app/models/dog_searcher.rb
@@ -71,7 +71,9 @@ class DogSearcher
   private
 
   def tracking_id_search?
-    @params[:search].to_i > 0 && @params[:search].to_i < 2_147_483_647
+    return false if @params[:search].to_s.match(/[A-z]/)
+    @params[:search].to_i > 0 &&
+    @params[:search].to_i < 2_147_483_647
   end
 
   def text_search?
--- a/app/models/dog_searcher.rb
+++ b/app/models/dog_searcher.rb
@@ -34,18 +34,10 @@ class DogSearcher
     @dogs = Dog.includes(:photos, :foster)
     if @manager
       @dogs = @dogs.includes(:adopters, :comments)
-      if tracking_id_search?
-        @dogs = @dogs.where(
-          'tracking_id = ? OR dogs.microchip = ?',
-          search_term,
-          search_term
-        )
-      elsif text_search?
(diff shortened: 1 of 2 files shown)


app/models/dog_searcher.rb
```diff
@@ -71,7 +71,9 @@ def self.search(params: {}, manager: false)
   private
 
   def tracking_id_search?
```
TheDudeWithTheThing: well what are we really trying to test here?  the whole string is a number?  The problem is that `.to_i` will parse for any number and return true now.

```
[1] pry(main)> "4 seasons".to_i
=> 4
```

maybe the real fix is to make sure the whole string is numbers

```
return @params[:search].to_s.match(/^[0-9]+$/)
```

http://rubular.com/r/Z89CWp7QdZ

TheDudeWithTheThing: Also I just learned that `===` when the regex is the first argument returns true / false, good stuff!

lnzd17: 👍 makes more sense to check for 'only numbers?' instead of checking for 'any letters?'. I will revise. Thanks!

markottaviani: I'll try and do a query of the production database tonight and see what the Microchip IDs actually look like.   I know several brands are in use and I kind of forget what the formats look like.  There is a chance that some are a blend of letters and numbers.

markottaviani: And of course now I see that people sometimes put spaces in the IDs...

lnzd17: haha of course. I will hold off with updates for now.

markottaviani: Okay, looks like letters sometime show up in the microchip values as well.   Maybe something like this for a different approach?  https://stackoverflow.com/questions/9379884/search-multiple-columns-rails

lnzd17: I added some changes considering the link above from @markottaviani.

lnzd17:
@markottaviani  Just wanted to check in on this PR and see if I should make any other changes. Thanks! :)

markottaviani:
Thanks for the reminder, I should have a chance to review this tonight.  @TheDudeWithTheThing LMK if you have any feedback.

TheDudeWithTheThing:
I'm good  👍

Issue #608, opened 2017-08-10
Add Translator to Staff Profile


markottaviani:
As a Volunteer I want to be able to note if I'm available to act as a Translator, and what languages I speak.  
As a Volunteer I want to be able to see if there are other volunteers available to act as a translator.  

Implementation:

Add a boolean and a text field to the User object for them to note if "I can translate from another language to English"  Yes/No   and then a text field where they can write in the languages other than English they speak.  

On the User Index View, add an additional Tab after "Foster Mentor" called "Translator".    Clicking on this tab should provide Name, Languages Spoken, Duties, and Location in the table below.

markottaviani:
@lnzd17 this should be fairly straight forward if you're looking for something to take on

PR #609, opened 2017-08-13
Translator #608

lnzd17:
PR for #608 
* User can now specify if they can translate English into one or more languages and they can list out those languages
* User #show page will indicate when a user has listed translation languages
* User #index page now has a translator filter tab

### DIFF
--- a/app/views/users/_user_table.html.erb
+++ b/app/views/users/_user_table.html.erb
@@ -3,12 +3,15 @@
     <thead>
       <tr>
         <th>Name</th>
+        <% if params[:translator] %>
+        <th> Known Languages </th>
+        <% end %>
         <th>Duties</th>
         <th>Phone</th>
         <th>Location</th>
         <% if params[:foster_mentor] == 'true' %>
           <th>Mentees</th>
-        <% else %>
+        <% elsif !params[:translator] %>
           <th>Fostered Dogs</th>
           <th>Coordinated Dogs</th>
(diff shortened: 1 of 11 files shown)


markottaviani:
Nice!  One tweak.  From the ```/users?translator=true``` index view, how about its show the following columns:

- Name
- Known Languages
- Duties 
- Phone
- Location

That way when looking for a translator for a specific language, one doesn't have to click on everyone's profile.  

Thanks!

lnzd17:
@markottaviani oops! Forgot that piece. Revised accordingly. 
With most recent changes in place, are all the appropriate fields for Foster Mentor still appearing?

app/views/users/_user_table.html.erb
```diff
@@ -3,12 +3,15 @@
...
+        <% if params[:translator] %>
+        <th> Known Languages </th>
+        <% end %>
         <th>Duties</th>
         <th>Phone</th>
         <th>Location</th>
         <% if params[:foster_mentor] == 'true' %>
           <th>Mentees</th>
-        <% else %>
+        <% elsif !params[:translator] %>
```
markottaviani: ```_user.html.erb``` needs a complimentary if statement as the current_foster_dogs and coordinated_dogs values still show up.  

https://cl.ly/0H21473G2d0j

lnzd17: 👍

Issue #573, opened 2017-05-05
Searchable Staff Resources


markottaviani:
>Please add to your very long list of items for the website, a search option for Staff Resources.  We have volunteers and fosters looking for items and not sure where they are.  Plus, we've moved things around and they are in new places!  Thank you!!!

Source: Mindy Email

TheDudeWithTheThing:
Looks like this is winning out in the poll about issues to work on. Is there any more info here?

What is a staff resource?
What is an item?
Moved real world objects or moved things on the web site?  If real world objects how are those being tracked today?

markottaviani:
This is the /folders section of the site.  Basically it's a hacked together file sharing system using paperclip gem.  Could easily benefit from a lot of love and improvement in the UI space.

lnzd17:
@markottaviani sorry for the delayed response. After ticket 608 I can take a crack at this one. Also still waiting to hear back on the adoption workflow ticket.

markottaviani:
Awesome thanks!

lnzd17:
@markottaviani  I am attempting to test out to current folder system on my local machine. I am about create a folder but not upload a file. I get the following message:

`missing region; use :region option or export region name to ENV['AWS_REGION']`

Does my environment need a secret variable?

markottaviani:
Sent you some ENV vars you can use.

lnzd17:
Update - working on adding a  search by filename on the folders#index

PR #610, opened 2017-08-19
Improve Staff Folder UI

lnzd17:
PR for #573

### DIFF
--- a/app/views/folders/_folder.html.erb
+++ b/app/views/folders/_folder.html.erb
@@ -1,2 +1,17 @@
-<h2><i class="fa fa-folder-open"></i> <%= link_to folder.name, folder  %>
-</h2><p><%= folder.description %></p>
+<tr>
+  <td>
+    <%= link_to folder_path(folder) do %>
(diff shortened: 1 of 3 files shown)


TheDudeWithTheThing:
This isn't directly related to the request about search. I wonder if we should just get this out separately from search, seems like a lot more info about resources which is probably helpful, then tackle the search.

lnzd17:
@markottaviani and @TheDudeWithTheThing I have created a table for the folders. It is ready to be merged...if we decide that this task is best completed on it's own and then later implement the search function ( I agree). 

Roughly how many folders and files are we dealing with? Does a user typically know the name of the file they are looking for and want to be able to search for the file on the folders#index page? Or would it be more helpful to implement a 'tagging' option and sort/filter by categories?

markottaviani:
@lnzd17 tagging is probably a good idea, and they may not know the name of the file they need.  

You can access the real deal by logging into your DMS account at https://ophrescue.org/folders

Also this table view looks awesome.  I think we just ship this for right now

PR #618, opened 2017-09-04
#573: Folder search

### LINKED ISSUE #573: Searchable Staff Resources

lnzd17:
fixes  #573 
* added search functionality for `attachment_name` and `attachment_description` on folders#index
* search returns `attachment_names` that match search parameters
* search will not return restricted attachments unless user has access

Need help with:
- implementing `joins` on `Attachment` and `Folder` instead of using  a `reject` loop on the `Attachment.where` query in `Folder_Searcher`

Need confirmation if the following items align with the issue:

-  Upon visiting folder#index a user will be presented with a table of all folders
A user has the option to search for attachments within the folders (search is done by file name or description)

-  Once a search is performed the results are returned on the same page (folder#index)

-  The search results are not folders but attachments and will be displayed in the same table where the folders were previously listed out.


markottaviani:
@lnzd17 Awesome!  Limiting to just the relevant files I think would be of big interest for folks.  

Playing around with the interface a bit, I think it's safe to totally ignore the folder and the folder's description, and just search the filenames and the filenames' descriptions.  

Search Return values could be:
* filename
* filename description
* last update timestamp of the file.

lnzd17:
Great! I will work towards that!

lnzd17:
@markottaviani search now only displays relevant files and I updated the column titles as you specified.

markottaviani:
Looks good!  Only issue is see right now when dealing with the "restricted" folders.  the file search right now doesn't check if the person has rights to the restricted folders or not.   So if a search turns one of those up, they'll be able to get to that.

lnzd17:
@markottaviani good catch! I will fix that and add tests.

lnzd17:
@markottaviani added tests and added logic for locked folders

markottaviani:
Looks good to me!  @TheDudeWithTheThing  any feedback?

app/controllers/folders_controller.rb
```diff
@@ -30,10 +30,13 @@ class FoldersController < ApplicationController
   before_action :admin_user, only: [:new, :create, :edit, :update, :destroy]
 
   def index
-    @folders = Folder.order(:name)
-
-    @folders.each do |a|
-      a.attachments.build
+    if params[:search]
```
TheDudeWithTheThing: while it fundamentally does the same thing I like to throw a `.present?` on these just so it reads better

app/models/folder_searcher.rb
```diff
@@ -0,0 +1,31 @@
...
+    new(user, params: params).search
+  end
+
+  private
+
+  def search_term
+    @params[:search]
+  end
+
+  def for_page(page = nil)
```
TheDudeWithTheThing: this is defined but never called

app/models/folder_searcher.rb
```diff
@@ -0,0 +1,31 @@
+class FolderSearcher
+
+  def initialize(user, params: {})
+    @user = user
+    @params = params
+  end
+
+  def search
+    query = 'attachment_file_name ILIKE :search OR description ILIKE :search'
+    @folder = Attachment.where(query, search: "%#{search_term.strip}%").map do |a|
```
TheDudeWithTheThing: you're assigning attachments to a var called folder, is this doing what you want?

Why not query unlocked resources only if a user can't download them, otherwise query all the things.

app/controllers/folders_controller.rb
lnzd17: sure i can add that!

app/models/folder_searcher.rb
lnzd17: @TheDudeWithTheThing I also felt odd about using *folder* for that assignment, however assigning it that way does help keep consistent when I want the view to show actual folders in the table versus when a search is made and the same table returns files (attachments). Do you foresee a problem with this approach going forward?

I may not be clear on your second point. From what I understand, I am returning only files that are appropriate based on the lock status of the file's parent folder and the current_user's permissions. Are you saying that this logic should not be inside the map loop?

TheDudeWithTheThing: how is a folder an attachment though?  I feel like I'm missing something.

if you were doing `a.folder` on line 12 then it would make sense

2nd point

You should be able to do this in the query

my quick version without validating anything works
```
@folder = Folder.joins(:attachments).where(query)
return @folder if @user.dl_locked_resources?

@folder = @folder.where(locked: false)
```

throw scopes on Folder for locked / unlocked and this is even simpler. Let me know if you have questions!

lnzd17: @TheDudeWithTheThing and @markottaviani 
I believe I understand the notes above, however let me outline how I interpreted this ticket as I may be at fault:
- Upon visiting `folder#index` a user will be presented with a table of all `folders`
- A user has the option to search for `attachments` within the folders (search is done by file name or description)
- Once a search is performed the results are returned on the same page (`folder#index`)
- The results are not `folders` but `attachments` and the same table that displayed the `folders`  before the search will now display the appropriate `attachments`

I agree that `@folders` within the `FolderSearcher` class is not appropriate and should be changed to `attachments`. That being said I wonder if it is best to change `FolderSearcher` to `AttachmentSearcher`.

 I have been attempting to use `joins` however I am running into an error that states:
`Can not eagerly load the polymorphic association`. However, I may not be building the query correctly.

This may be best discussed during our meeting, but wanted to add my thoughts here as well. Thanks for helping me out with this! :)

app/models/folder_searcher.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def initialize(user, params: {})
+    @user = user
+    @params = params
+  end
+
+  def search
+    query = 'attachment_file_name ILIKE :search OR description ILIKE :search'
+    search = "%#{search_term.strip}%"
+    attachment = Attachment.where(query, search: search ).map do |a|
```
tobymurray: Is this iterating over the result and removing locked resources? Perhaps `select` is more direct than `map` and `compact`?

app/views/folders/index.html.erb
```diff
@@ -5,12 +5,22 @@
   <%= link_to "New Folder", new_folder_path, class: "btn btn-primary" %>
   </p>
 <% end %>
+<%= link_to "All Folders", folders_path, class: 'btn btn-primary'%>
+<br/>
+<br/>
+<%= form_tag folders_path, :method => 'get', :class => 'form-inline' do %>
```
tobymurray: Minor and purely style, but if I recall correctly I think the "hashrocket" has fallen out of fashion, with a reasonably complete replacement as of Ruby 2.2 with the JSON style hash notation - e.g. `method: 'get', class: 'form-inline'`, although I don't know if there are still places you need the hashrocket...

markottaviani: Yup, generally for new stuff use the updated hash notation unless something breaks :)

Running rubocop is a great way to pick on these things too when you're working on code.  I'm try to do a better job of running it against specific files when I'm working on them.

lnzd17: @markottaviani and @tobymurray Will update :)

TheDudeWithTheThing: hash rocket for non symbols basically

app/models/folder_searcher.rb
TheDudeWithTheThing: Oh no we forgot to discuss! I can look at this today.

PR #623, opened 2017-09-14
added has_family_under_18 boolean field

blai00:


markottaviani:
Did you check out the ```rspec``` test results?

blai00:
Rspec showed the same results as the master branch.

markottaviani:
so in other words your tests for this branch pass fine locally?

spec/features/apply_to_adopt_spec.rb
```diff
@@ -213,6 +214,7 @@
     fill_in('adopter_adoption_app_attributes_how_did_you_hear', with: 'Google and Craigslist')
     fill_in('adopter_adoption_app_attributes_ready_to_adopt_dt', with: '2012-12-13')
     choose('adopter_adoption_app_attributes_is_ofage_true')
+    choose('adopter_adoption_app_attributes_has_family_under_18')
```
markottaviani: This needs to be ``` choose('adopter_adoption_app_attributes_has_family_under_18_true')```

blai00:
I always get these errors, I am not sure why but am I missing something? 
screen shot 2017-09-14 at 5 36 57 pm

Will push that change.

markottaviani:
Something missing for sure, since the tests should be green for you locally (though you can see they're finally green on travis!)   Can you provide the full rspec output?

blai00:
I have the full coverage report, but I don't think I can take a screenshot of the whole report because it cuts out. Will that suffice?

markottaviani:
No need for screenshot, highlight and copy paste the text, or have rspec dump the output to a text file and attach it to the pull request.

blai00:
I attached the rspec file.
[rspec.txt](https://github.com/ophrescue/RescueRails/files/1305055/rspec.txt)

markottaviani:
Ah! Lots of errors saying that you're missing phantomjs.

I'm so used to having it installed it never made it into the readme! I'll make sure to fix that.  

See this page:  https://github.com/teampoltergeist/poltergeist/blob/master/README.md for instructions on how to install Phantomjs.  After you do that the tests should pass.  If not, send the full rspec dump again.

blai00:
Awesome, it worked! I have 0 failures now.

spec/factories/adoption_app.rb
```diff
@@ -24,6 +24,7 @@
     vet_info 'Dr. Spaceman Baltimore MD'
     max_hrs_alone 3
     is_ofage true
+    has_family_under_18 true
```
TheDudeWithTheThing: if our default is nil then why give this a value of true in the factory?

db/migrate/20170914220910_add_has_family_under18_to_adoption_app.rb
```diff
@@ -0,0 +1,5 @@
+class AddHasFamilyUnder18ToAdoptionApp < ActiveRecord::Migration[5.0]
```
TheDudeWithTheThing: We've been bad about this, but migrations adding columns should be deployed by themselves so that the model isn't looking for a column that isn't in the DB yet.  That's up to @markottaviani though since he does the deploys.

spec/factories/adoption_app.rb
blai00: Should it be nil then in the factory? I am not familiar with the use of this and I just thought that I was just testing it for the purpose of seeing if it would work?

db/migrate/20170914220910_add_has_family_under18_to_adoption_app.rb
blai00: Just curious since I am new to all of this, how would that work? Do we just not do this locally and deploy?

markottaviani: @TheDudeWithTheThing is referring to the production deployments.    I believe with the way I have the Capistrano stuff setup the database migrations happen first anyway. However I agree with trying to split out the data model ahead of time as much as we can.

spec/factories/adoption_app.rb
markottaviani: @blai00 if you've done very little with RSpec testing, consider checking out https://leanpub.com/everydayrailsrspec

app/views/adopters/_adopter_form.html.erb
```diff
@@ -2,6 +2,9 @@
   <table class="table table-striped table-bordered table-condensed">
     <tr>
       <td colspan="2">
+        <% if @adopter.adoption_app.has_family_under_18 == false %>
```
markottaviani: This should be ```true```

spec/features/apply_to_adopt_spec.rb
```diff
@@ -213,6 +214,7 @@
     fill_in('adopter_adoption_app_attributes_how_did_you_hear', with: 'Google and Craigslist')
     fill_in('adopter_adoption_app_attributes_ready_to_adopt_dt', with: '2012-12-13')
     choose('adopter_adoption_app_attributes_is_ofage_true')
+    choose('adopter_adoption_app_attributes_has_family_under_18_true')
     choose('adopter_adoption_app_attributes_house_type_own')
```
markottaviani: Add a ```expect(page).to have_content('Has Family Under 18')```  toward the bottom of the test to verify that the value is showing up in the application view correctly.

blai00:
Awesome. I will look at waffle to see what I can do next. Anything that you think would be best for me to tackle?

Issue #437, opened 2016-02-27
Users Staff Directory Should Support Canadian Mailing Addresses


markottaviani:
```/users``` lacks a Country Field Dropdown.  

@tobymurray might be a good start for you so we can properly add you to the volunteer directory! 😄

tobymurray:
As in CRUDing the users is missing the field? Limit it to any specific countries, or pull in everything?

markottaviani:
Yup!  

Limit to United States of America and Canada is good for now (default to US in the database migration) .  

Break out the database migration into a separate pull request, because we like to deploy those our first.  

Also, if you see any other issues with the input validation for Users Object's address field to properly support Canada, note those here and we'll try to sort them out.  Thanks!

tobymurray:
Canada has provinces/territories instead of states (theoretically could abuse the current field), and 6 character alphanumeric postal codes instead of zip codes (oddly enough compatible with the database, but not the User model as written). So definitely a little work to do there to support it nicely. I think a localized form would be the most useful way of handling it, I'm sure it would be straightforward enough with a little JS, and there's likely one on the internet somewhere we could leverage.

It's also looking like it'd be nice to break out the addresses into a separate table, so multinational support would be free for `adopters` and `banned_adopters`. Potentially `events` as well? Not saying it should be done as part of this, but something to think about.

tobymurray:
Fleshing out what the application layer refactoring looks like to support [name change PR](https://github.com/ophrescue/RescueRails/pull/629), I pushed a branch [here](https://github.com/ophrescue/RescueRails/compare/437-Use-more-generic-address-terms...437-Rename-application-layer-properties?expand=1). My intention is that absolutely nothing is different from the user's perspective; the API changes (I assume it only drives the UI?) and the `User` model property names change. 

From there, I was thinking:
- Add application layer support for province and non-US postal code
- Add `country` property to user
- Update `User` form to toggle between state/ZIP and province/postal code depending on which country is selected

Does this sound like it lines up with what was in mind for this ticket?

PR #627, opened 2017-09-20
#437: Add 'country' to user as 3 character string

tobymurray:
Looking at the schema, it looks like state is either a 2 letter code, as in `banned_adopters`, or a string, as in `adopters` and `users`. Using letter codes is a nice way to have a canonical representation in the database and maintain readability. I've added country as an optional 3 letter code - personally, I find 3 letter country codes easier to read (CAN doesn't look like California). I used the ISO 3166-1 alpha-3 for the US as the value for all legacy data (see comment [here](https://github.com/ophrescue/RescueRails/issues/437#issuecomment-330719943)).

I added a comment on the column, acknowledging that this is inconsistent with the current state. It manifests in both the `schema.rb` and also in the database itself, e.g.:
```
# \d+ users
                                                                                Table "public.users"
            Column            |              Type              |                     Modifiers                      | Storage  | Stats target |             Description              
------------------------------+--------------------------------+----------------------------------------------------+----------+--------------+--------------------------------------
... 
 country                      | character varying(3)           |                                                    | extended |              | Country as a ISO 3166-1 alpha-3 code

### DIFF
--- a/db/migrate/20170920014235_add_country_to_user.rb
+++ b/db/migrate/20170920014235_add_country_to_user.rb
@@ -1,0 +1,10 @@
+class AddCountryToUser < ActiveRecord::Migration[5.0]
+  def up
+    add_column :users, :country, :string, limit: 3, comment: 'Country as a ISO 3166-1 alpha-3 code'
+    execute "UPDATE users SET country = 'USA';"
+  end
+
+  def down
+    remove_column :users, :country
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


markottaviani:
LGTM.   @TheDudeWithTheThing any thoughts on this?   Will hold off on the merge + deploy until after the rest of the code is ready to go and will ship them out one after the other, cuts down on any nulls sneaking in.

tobymurray:
I'll resubmit this when it's closer to landing. I think a 2 character code will be easier and more conventional, albeit less readable...

markottaviani:
I'm Pro 3 Digit County Code

tobymurray:
Yeah? Works for me. I like 3 character codes more, looking around it seems like it's slightly less conventional. Both are standards though, so no issues.

PR #628, opened 2017-09-20
Event validation

markottaviani:
Some extra Event Validation to cut down on errors caused by people pasting in HUGE URLS.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -52,13 +52,15 @@ class Event < ApplicationRecord
                         :address,
                         :description
 
-  validates_format_of :location_url, with: URI::regexp(%w(http https))
+  validates_format_of :location_url, with: URI::regexp(%w[http https])
 
   validates :title, length: { maximum: 255 }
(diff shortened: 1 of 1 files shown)


app/models/event.rb
```diff
@@ -52,22 +52,31 @@ class Event < ApplicationRecord
...
+  validates_format_of :location_url, with: URI::regexp(%w[http https])
+
+  validates :title, length: { maximum: 255 }
+  validates :location_name, length: { maximum: 255 }
+  validates :address, length: { maximum: 255 }
+  validates :location_url, length: { maximum: 255 }
+  validates :facebook_url,
+            length: { maximum: 255 },
+            allow_blank: true,
+            presence: { message: 'Remove everything after the ? (You just need https://facebooks.com/events/12345/)' }
```
tobymurray: Is this intentionally `facebooks`?

markottaviani: Good catch, thanks!

PR #629, opened 2017-09-23
#437: Change 'state' to 'region' and 'zip' to 'postal code'

tobymurray:
I wanted to see if this makes sense or not before going too far with it - I took a look at what refactoring would be required to fully flesh this out and it's extremely doable so far as I can tell.

**Motivation:**

The desire is to move from supporting exclusively US mailing addresses to supporting international addresses, starting with Canada. Countries are not consistent with how they name their administrative divisions. "Region" appears to be a conventional term for the principal subdivisions of a country - whether they're states, provinces, or other. As a single reference point, see http://schema.org/PostalAddress

Additionally, a ZIP code is a particular type of postal code, but the terminology "ZIP code" is specific to the US, so this changes to use a more generic term.

I find it's generally valuable to make sure the database column names line up with the data that's inside them, otherwise code starts to look crazy quickly, so this seems like a good move with regards to supporting international addresses.

**Concerns:**
- Addresses are not normalized, so the same change would likely be desirable for other representations of the addresses if it is desirable for this one.
- The `User` model has a validation that states must be strictly 2 uppercase characters -  if there was legacy data at some point before this existed or what have you, reducing the field limit like this is "dangerous". I added the `select` at the beginning as a sanity check to make sure the operation is safe - definitely different ways that could be handled though.

### DIFF
--- a/db/migrate/20170923045153_change_user_address_column_names.rb
+++ b/db/migrate/20170923045153_change_user_address_column_names.rb
@@ -1,0 +1,22 @@
+class ChangeUserAddressColumnNames < ActiveRecord::Migration[5.0] 
+  def up
+    incompatible_records = execute("SELECT id, name, email, state FROM users WHERE LENGTH(state) > 2;")
+    raise "Migration aborted beacuse legacy data contains users with states that are more than two letters: #{incompatible_records.to_a}" unless incompatible_records.to_a.empty?
+    
+    rename_column :users, :state, :region 
+    change_column :users, :region, :string, limit: 2, comment: 'Region (state or province) as a 2 character ISO 3166-2 code' 
+ 
+    rename_column :users, :zip, :postal_code 
+    change_column_comment :users, :postal_code, 'Postal code - ZIP code for US addresses' 
+  end 
+ 
+  def down 
+    rename_column :users, :region, :state 
+    change_column :users, :state, :string, limit: 255, comment: '' 
+ 
+    rename_column :users, :postal_code, :zip 
+    
+    # Remove inapplicable comment
+    change_column_comment :users, :zip, '' 
+  end 
+end 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -81,6 +81,9 @@ require 'digest'
 class User < ApplicationRecord
   include Filterable
 
+  alias_attribute :state, :region
+  alias_attribute :zip, :postal_code
+
   attr_accessor :password,
                 :accessible
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -308,8 +308,8 @@ ActiveRecord::Schema.define(version: 20170923045153) do
     t.string   "address1",                     limit: 255
     t.string   "address2",                     limit: 255
     t.string   "city",                         limit: 255
-    t.string   "region",                       limit: 255,                                            comment: "Postal region - state or province"
-    t.string   "postal_code",                  limit: 255,                                            comment: "Postal code - ZIP codes for US addresses"
+    t.string   "region",                       limit: 2,                                              comment: "Region (state or province) as a 2 character ISO 3166-2 code"
+    t.string   "postal_code",                  limit: 255,                                            comment: "Postal code - ZIP code for US addresses"
     t.string   "duties",                       limit: 255
     t.boolean  "edit_dogs",                                              default: false
     t.text     "share_info"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170917020834) do
+ActiveRecord::Schema.define(version: 20170923045153) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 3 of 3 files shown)


markottaviani:
LGTM, summoning @TheDudeWithTheThing

db/migrate/20170923045153_change_user_address_column_names.rb
```diff
@@ -0,0 +1,22 @@
+class ChangeUserAddressColumnNames < ActiveRecord::Migration[5.0] 
+  def up
+    incompatible_records = execute("SELECT id, name, email, state FROM users WHERE LENGTH(state) > 2;")
+    raise "Migration aborted beacuse legacy data contains users with states that are more than two letters: #{incompatible_records.to_a}" unless incompatible_records.to_a.empty?
```
TheDudeWithTheThing: this seems weird to just bomb out, @markottaviani can we get a dump or something to clean these up before this runs?

Also why are you doing raw sql here?  You can do some fancy stuff instead

`User.where('LENGTH(state) > 2').exists?`

tobymurray: > this seems weird to just bomb out

I don't know how you guys typically handle data migrations, so this is only here to make it abundantly clear that this could theoretically be dangerous (destroy production data). I thought it was safer for the migration to fail under the described condition. If the data is checked before migrating, then it's not technically necessary.

> can we get a dump or something to clean these up before this runs

My intention was that this query would provide enough information to be able to identify any users whose state violates this constraint (if there are any). 

> Also why are you doing raw sql here? You can do some fancy stuff instead

The short answer is because I knew how to get the result I wanted faster with SQL than with ActiveRecord. 

Do you prefer "fancy stuff"? Introducing models into migrations has been something that has been non-trivial in my experience, as the `User` model is not necessarily defined in the way you'd expect when this migration runs. The guide touches on this - see [here](http://guides.rubyonrails.org/v3.2/migrations.html#using-models-in-your-migrations)

>When creating or updating data in a migration it is often tempting to use one of your models. After all, they exist to provide easy access to the underlying data. This can be done, but some caution should be observed.

Some people re-declare the models in the migrations to address this, which is another complication. In this specific case I don't imagine it matters one way or another. I've generally found it's easier to default to avoiding models where they're not necessary, as in my view there's nothing wrong or undesirable about SQL. 

All that to say, if the SQL is simple is there a downside to it/is it less desirable for you?

TheDudeWithTheThing: SQL should be fine.  Fancy stuff is good too. We run a data migration gem or would make a rake task to do something like this since it's a 1 off and not really important to be part of the db schema definition. In those you'd have access to the models.

Careful with that doc link you posted, it's for Rails 3.2, we're on 5.

tobymurray: I'm not clear on the outcome you'd like based on what you said - are you wanting the `select` removed from this file, as it'll be handled outside the migrations? It's not part of the schema definition as presently implemented. 

> Careful with that doc link you posted, it's for Rails 3.2, we're on 5.

Yeah, that was the first result from a google search. As a resource for that particular concern it's still valid, as far as I know nothing has changed in that department. If you're looking for more info on the issue more recent sources may be more useful.

TheDudeWithTheThing: I should also note that I apologize if I’m talking down to you, no clue what your experience level is, looking forward to getting to know that at our meeting soon!

I think for this we want to separate the renaming of the column and the fixing of the data. I would say that the steps here are

1. Prevent anything bigger than 2 characters into state with a rails model validation.
2. See if any data needs fixing, if it does fix it with a rake task or direct sql.
3. Put constraints on the column.
4. Rename column.

3 and 4 can be the same migration.

What happens after the column rename to the alias_attribute now that you are aliasing a column that no longer exists?

We could just do without that as the migration would be quick and we haven’t been going for 0 downtime migrations. If we were you would do something like https://github.com/ankane/strong_migrations/blob/master/README.md#renaming-or-changing-the-type-of-a-column.

markottaviani:
@tobymurray @TheDudeWithTheThing  thank you both for the diligent analysis on the database migration.  Looking forward to getting everyone together next week!

#DOGCON2017

Happy to report that 
```SELECT id, name, email, state FROM users WHERE LENGTH(state) > 2;```
Returns zero results in the Prod database.  

We additionally do have a ```validates :state, length: { is: 2 }``` in User to keep anything ugly from getting in.  

Once Upon a time there was ugly data in the state field.  I added the validation, then did a database cleanup, but never modified the actual database column.

db/migrate/20170923045153_change_user_address_column_names.rb
tobymurray: No stress! I appreciate you being as thorough as possible, as it will help me with on-boarding. I didn't assume you were talking up or down. I hope I didn't come across as offended, I'm making an effort to explain my reasoning behind everything to try and short circuit any confusion.

> Prevent anything bigger than 2 characters into state with a rails model validation.
@markottaviani referenced that this is already in place on the `User` model

> See if any data needs fixing, if it does fix it with a rake task or direct sql.
Sounds like that's all good as well

> Put constraints on the column.
Rename column.

Sounds good

> What happens before the column rename to the alias_attribute when it is referencing a column that doesn’t exist??

You mean if the `User` model is deployed and the server is started before the database is migrated? I imagine almost all workflows would break, as you'd end up the same as any other missing attribute, e.g. `NoMethodError: undefined method 'region=' for #<User:0x00000008043cf8>` for the setter and the same for the getter.

tobymurray:
Should I merge this? Do you guys merge when you're ready?

db/migrate/20170923045153_change_user_address_column_names.rb
markottaviani: Our capistrano setup migrates the database changes first

PR #633, opened 2017-09-29
#437: Change User 'state' to 'region' and 'zip' to 'postal code'

tobymurray:
Application layer refactor that corresponds to [this database migration](https://github.com/ophrescue/RescueRails/pull/629). The UI is unchanged, so now the field labeled `state` corresponds to the attribute `region` and the field labeled `zip` corresponds to the attribute `postal_code`. This is to support upcoming work that will conditionally change the UI to show localized labels - "Province" and "Postal Code" for Canadian addresses.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -31,8 +31,8 @@
 #  address1                     :string(255)
 #  address2                     :string(255)
 #  city                         :string(255)
-#  state                        :string(255)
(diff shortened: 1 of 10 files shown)


tobymurray:
@TheDudeWithTheThing or @markottaviani - can you take a look at this when you have a chance?

markottaviani:
Looks good to me, @TheDudeWithTheThing  ?

PR #635, opened 2017-10-03
More adopter validations

markottaviani:
Adding some additional server side validations since we've been getting some weird spam garbage submissions hitting the error logs.

### DIFF
--- a/app/models/adopter.rb
+++ b/app/models/adopter.rb
@@ -12,7 +12,6 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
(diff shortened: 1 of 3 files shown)


app/models/adopter.rb
```diff
@@ -73,10 +72,6 @@ class Adopter < ApplicationRecord
 
   FLAGS = ['High', 'Low', 'On Hold']
 
-  def attributes_to_audit
```
markottaviani: This was a duplicate method 🤦‍♂️

PR #638, opened 2017-10-06
575 new user tabs

blai00:


db/schema.rb
```diff
@@ -256,6 +256,19 @@
     t.boolean  "locked",                  default: false
   end
 
+  create_table "fosters", force: :cascade do |t|
```
markottaviani: Where does this Foster table keep coming from?

blai00: I am not sure but anytime I see that the migrations don't work or whatever is going on is really weird, I clone a clean copy to make these changes. I have a feeling this is how I setup or seed the database? I am not entirely sure though.

PR #640, opened 2017-10-07
fixes #639

### LINKED ISSUE #639: Folder Search Error
```
A NoMethodError occurred in folders#index:

undefined method `locked?' for nil:NilClass
app/models/folder_searcher.rb:28:in `locked_folder?'
```

lnzd17:
fixes #639
Updated `folder_searcher` to only return `Attachments` that have the `attachable_type` of `Folder`.

Does this solution cause too many queries to the DB?


markottaviani:
Probably okay, but will keep an eye on NewRelic if there is anything weird.

TheDudeWithTheThing:
I still think we can get it done in 1 query, we have the know how!

PR #642, opened 2017-10-09
Hotfix/schema and state fixes

markottaviani:
Found some holes in our test coverage for some of the views that use state.   Cleaned up the views and added a few more tests.


TheDudeWithTheThing:
a foster isn't a user then?  weird that this isn't working

`  alias_attribute :state, :region`

markottaviani:
A foster is a user

tobymurray:
Yikes, my bad. That looks like a decent amount of broken stuff... Thanks for the quick fix, I'll have to be a bit more careful next time.

markottaviani:
No worries! We should have had better tests!

PR #643, opened 2017-10-10
437 add country to user model

tobymurray:
- Adds the `country` column to `User` model
-- Adds value for the user in `seed.rb`, as well as the user factory
- Sets all legacy users to have country value of `USA`
- Defaults any new/edited users that don't otherwise have a value for the `country` field to be `USA`

This is an effort to keep the migration separate from most of the application layer work - the next step will be fleshing out `country`, adding validations, and adding the field in the UI.


app/models/user.rb
```diff
@@ -251,4 +253,8 @@ def make_salt
     def secure_hash(string)
       Digest::SHA2.hexdigest(string)
     end
+
+    def default_country_to_usa
+      self.country ||= "USA"
```
TheDudeWithTheThing: you're doing this in the DB already, why do it in the app too?

tobymurray: I set all legacy users to have the country `USA` in the database, but didn't set a default. This handles new/edited users until such a time as it's specified on creation/editing.

tobymurray:
>  Request to 'http://127.0.0.1:43103/' failed to reach server, check DNS and/or server status - Timed out with the following resources still waiting https://www.facebook.com/impression.php/fb40b16/?lid=115&payload=%7B%22source%22%3A%22jssdk%22%7D

That's an odd failure. Is some network traffic still taking place during specs?

markottaviani:
Glitch in the Matrix with the Facebook JS stuff on some of the pages I guess.  I triggered a rerun of the build and we went green.

PR #645, opened 2017-10-12
Use rails simplecov configuration

tobymurray:
This has two separate changes. The simple change is moving the `SimpleCov` initialization code from `rails_helper` to `spec_helper` - `SimpleCov` needs to be `require`'d before any other application code, or it won't be able to track coverage on that code. Practically speaking, this doesn't look like it makes any difference whatsoever with the current implementation, but as `spec_helper` is `require`'d before `rails_helper`, the initialization code should be in `spec_helper`.

Secondly, it looks like we currently use the default configuration of SimpleCov. There's a different configuration for Rails projects that tries to make some sensible decisions about what to test and what not to test. See [documentation](https://github.com/colszowka/simplecov#profiles) for more info. 

I was hoping Coveralls would show a nice view of what had actually changed, but that doesn't look like the case.

What's not included now in the calculation of code coverage (can be added if it's valuable):
- spec files (I don't think this makes sense)
- spec factory classes
- everything in the `config` directory

What's included now in the calculation of code coverage (can be removed if not valuable):
- ruby code in the `lib` directory

Overall, I think this is a more accurate representation of the code coverage, as it's not skewed by all the `100%` coverage on spec files.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -1,12 +1,3 @@
-require 'simplecov'
-require 'coveralls'
-
-SimpleCov.formatters = [
-  SimpleCov::Formatter::HTMLFormatter,
-  Coveralls::SimpleCov::Formatter
-]
-SimpleCov.start
-
 ENV['RAILS_ENV'] ||= 'test'
 require 'spec_helper'
 require File.expand_path('../../config/environment', __FILE__)
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,3 +1,12 @@
+require 'simplecov'
+require 'coveralls'
+
+SimpleCov.formatters = [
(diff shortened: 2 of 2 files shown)


TheDudeWithTheThing:
Is the drop in coverage to 75% from this? 

Good change for sure and we can go more crazy with filtering out files or other dirs.

tobymurray:
Yeah, the drop in coverage reflects that this is now running on 1327 lines of code whereas before we were running on 2754 lines of code. The majority of those now-uncovered 1427 lines of code were "100% covered" spec files.

Fundamentally, I don't think percentage covered is a particularly good metric. Really we're interested in the number of lines of code that are not covered. That metric would not have changed nearly as much with this modification.

markottaviani:
Thanks for the ice cold bucket of reality!

PR #647, opened 2017-10-15
437: Validate user country

tobymurray:
Adds validation that the country is either `USA` or `CAN`. The behavior to default all users to `USA` if it's not otherwise specified is still in place. Functionally I don't expect this to change anything, it's just another step closer to supporting two different countries.


markottaviani:
LGTM  @TheDudeWithTheThing   any comments?

PR #650, opened 2017-10-19
Ranks dogs with query at start of name before others

TheDudeWithTheThing:


TheDudeWithTheThing:
@markottaviani the sql injection is real, let me see if I can figure out how to sanitize this. Since it's an order by you can't just do `order('?', value)`

PR #651, opened 2017-10-20
Added ranking and sorting tests and fix regression

markottaviani:
This switches us back to the old sorting methods, and as the fix still didn't sort correctly and also broke the by column sorting.  I moved the code around a little bit so it's easier to see what's going on.  I also added new tests to very at the column by column sorting works.

### DIFF
--- a/app/models/dog_searcher.rb
+++ b/app/models/dog_searcher.rb
@@ -95,7 +95,7 @@ class DogSearcher
   end
 
   def sort_direction
-    %w[asc desc].include?(@params[:direction]) ? @params[:direction] : 'desc'
+    %w[asc desc].include?(@params[:direction]) ? @params[:direction] : 'asc'
   end
 
   def for_page(page = nil)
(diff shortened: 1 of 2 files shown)


app/models/dog_searcher.rb
```diff
@@ -77,14 +77,25 @@ def with_includes
...
   def with_sorting
-    sort = if text_search?
-      sort_string = "case when name ilike ? then 1 else 2 end, ? ?"
-      [sort_string, search_term + '%', sort_column, sort_direction]
-    else
-      ['? ?', sort_column, sort_direction]
-    end
+    @dogs = @dogs.order(sort_column + ' ' + sort_direction)
+
+    # Falling back to the old sort stuff here for now
```
tobymurray: Given that this is in git, is it useful to keep the commented out code without explanation for why it's commented out?

TheDudeWithTheThing:
What happened? Sorry I should have mentioned I wasn’t fully done yet.

markottaviani:
All good.  The merge broke the sort by clicking on a column header.  We didn't have tests for that so this adds them.

TheDudeWithTheThing:
omg so stupid, I didn't assign the sorted results back to `@dogs`, how the hell did that ever work?

PR #652, opened 2017-10-20
Fixes the ranking for real

TheDudeWithTheThing:
What a n00b

What did we learn here?

* to pass in bind params to order by you pass in an array
* bind params will be quoted properly, so don’t do that for normal sorts
* assign results back to the var, don't be a n00b

### DIFF
--- a/app/models/dog_searcher.rb
+++ b/app/models/dog_searcher.rb
@@ -77,19 +77,14 @@ class DogSearcher
   end
 
   def with_sorting
-    @dogs = @dogs.order(sort_column + ' ' + sort_direction)
+    sort = "#{sort_column} #{sort_direction}"
 
-    #
-    # Progress on #649
-    #
-    # sort = if text_search?
-    #   sort_string = "case when name ilike ? then 1 else 2 end, ? ?"
-    #   [sort_string, search_term + '%', sort_column, sort_direction]
-    # else
-    #   ['? ?', sort_column, sort_direction]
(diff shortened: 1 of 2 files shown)


TheDudeWithTheThing:
Did we have that sql injection warning before?  It's sanitized with valid values so not sure why it's complaining.

markottaviani:
It didn't flag on the old branch.

TheDudeWithTheThing:
ha

> Confidence level: Weak

TheDudeWithTheThing:
It must be flagging now because of the couple of layers involved of string interpolation here. Is there no way to mark it as ok?

markottaviani:
Yeah we can flag it as false positive in Hakiri

markottaviani:
I fixed up the rspec tests for this.  The initial search is working correctly, however if someone attempts to sort the results instead of accepting the default sort, the results aren't honored.  The respec tests now break to reflect this, and here is a GIF that shows the behavior as well. 

https://cl.ly/271S0p410Y0r

TheDudeWithTheThing:
What’s the behavior we want? tt will be first always in the current state, only do that when name is the sort column?

markottaviani:
I think the desired behavior is:

When a text search is performed without specifying any sort values, tt is first.

With a text search is performed with sort values, honor the standard sorts.  (don't rank TT first)

markottaviani:
false positive cleared

PR #653, opened 2017-10-20
Add region validator

tobymurray:
Continuing the saga of support for Canadian addresses, this introduces validation on the region (state/province) when the country is Canada. The region code is validated to be 2 letters regardless, but when the user has a Canadian address the region is additionally checked to ensure it's actually a province.

### DIFF
--- a/app/validators/region_validator.rb
+++ b/app/validators/region_validator.rb
@@ -1,0 +1,43 @@
+class RegionValidator < ActiveModel::Validator
+  PROVINCES = %w[AB BC MB NB NL NS ON PE QC SK NT NU YT].freeze
+
+  def validate(record)
+    country = ISO3166::Country.find_country_by_alpha3(record.country)
+
+    unless country
+      record.errors[:region] << "cannot be validated without a value for country."
+      return
+    end
+
+    if country.eql? ISO3166::Country[:us]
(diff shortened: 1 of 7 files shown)


app/validators/region_validator.rb
```diff
@@ -0,0 +1,43 @@
+class RegionValidator < ActiveModel::Validator
+  PROVINCES = %w[AB BC MB NB NL NS ON PE QC SK NT NU YT].freeze
+
+  def validate(record)
+    country = ISO3166::Country.find_country_by_alpha3(record.country)
+
+    unless country
```
TheDudeWithTheThing: `if country.blank?` ?

tobymurray: `find_country_by_alpha3` returns either a `ISO3166::Country` or `nil`, so I think `blank?` is unnecessary (and because of that a little misleading). I'd be happy with `if country.nil?`, if that would be preferable?

TheDudeWithTheThing: yeah I go with if over unless when possible. I also like to be explicit with what you’re checking for so nil? would work.

tobymurray: Works for me! Swapped out for `nil?` in 02e3367

PR #655, opened 2017-10-20
#654: Fix user ZIP code validation regex

### LINKED ISSUE #654: User ZIP validation allows invalid ZIP codes
Presently the validation for a user's ZIP code is:
```
 validates_format_of :postal_code, 
                    with: /\A\d{5}-\d{4}|\A\d{5}\z/, 
                    message: "should be 12345 or 12345-1234", 
                    allow_blank: true 
```
As I understand it ZIP codes can either be 5 digits or 9 digit ZIP+4. The regex is missing the terminating character (`\z`) in the first alternative, which means the regex that is intended to match 4 digits matches anything starting with 4 digits. It looks like it's intended to be:
`/\A\d{5}-\d{4}\z|\A\d{5}\z/`

An example of an inappropriate match can be illustrated with, e.g.:
`User.new(name: "a", email: "a@b.c", country: "USA", region: "ON", postal_code: "12345-1234-Not-a-valid-postal-code...").save!`

Perhaps it'd be slightly more clear to use an optional instead of alternate, e.g. `\A\d{5}(?:-\d{4})?\z`?

ZIP codes that have slipped through the cracks can likely be found with a query, e.g.:
`SELECT id, name, email, postal_code FROM users WHERE postal_code !~ '\A\Z|\A\d{5}(?:-\d{4})?\Z';`

tobymurray:
Before this fix is merged, the production database should be checked for invalid data. For example, anything returned by:
`SELECT id, name, email, postal_code FROM users WHERE postal_code !~ '\A\Z|\A\d{5}(?:-\d{4})?\Z';`
will be invalid after this has been merged.

Fixes #654. The regular expression used was missing the terminating character in the first alternative expression. This means that any value could be used for the ZIP so long as it started with 5 digits followed by 4 digits.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -148,17 +148,17 @@ describe User do
     context 'with invalid fields' do
       it 'is invalid with a state more than 2 letters' do
         user = build(:user, region: 'Penn')
-        user.valid?
+        expect(user).not_to be_valid
         expect(user.errors[:region]).to include('is the wrong length (should be 2 characters)')
       end
       it 'is invalid with a zip code of more than 5 characters' do
         user = build(:user, postal_code: 'virgina')
-        user.valid?
+        expect(user).not_to be_valid
         expect(user.errors[:postal_code]).to include('should be 12345 or 12345-1234')
       end
-      it 'is invalid with a zip code that starts valid but contains more digits' do
+      it 'is invalid when it starts with a valid zip code but contains extra characters' do
         user = build(:user, postal_code: '12345-12345')
-        user.valid?
+        expect(user).not_to be_valid
         expect(user.errors[:postal_code]).to include('should be 12345 or 12345-1234')
       end
     end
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -156,6 +156,11 @@
         user.valid?
         expect(user.errors[:postal_code]).to include('should be 12345 or 12345-1234')
       end
+      it 'is invalid with a zip code that starts valid but contains more digits' do
+        user = build(:user, postal_code: '12345-12345')
+        user.valid?
```
TheDudeWithTheThing: `expect(user).not_to be_valid`

TheDudeWithTheThing: probably change the other 2 above as well

tobymurray: Can do - did so in dae7f03

markottaviani:
Prod Database verified and cleaned up!

PR #657, opened 2017-10-21
437: Add validator for postal code

tobymurray:
As the final piece of Canadian address validation, this introduces a postal code validator for both US ZIP codes and Canadian postal codes. Validation remains the same as before for US addresses (with the quirk that it can't be validated if the country isn't specified), and it validates only that it's the right form for Canadian postal codes. I didn't look at all at code to check it's a real postal code.


tobymurray:
Hmm, I introduced a bug with this - validating against a different property than what is written in the database. I'll clean it up.

PR #661, opened 2017-10-26
#437: Sanitize postal code before validation

tobymurray:
Fixes bug I accidentally introduced in [this PR](https://github.com/ophrescue/RescueRails/pull/657). The validation I introduced was operating on a manipulated version of the model's attribute. This sanitizes the postal code (removes whitespace and converts letters to upper case) before validation.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -258,8 +258,8 @@ class User < ApplicationRecord
     end
 
     def sanitize_postal_code
-      return unless self.postal_code
-      self.postal_code = self.postal_code.delete(' ').upcase
+      return if postal_code.blank?
+      self.postal_code = postal_code.delete(' ').upcase
     end
 
     def country_is_supported
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -137,6 +137,7 @@ class User < ApplicationRecord
   belongs_to :mentor, class_name: 'User', foreign_key: 'mentor_id'
   has_many :mentees, class_name: 'User', foreign_key: 'mentor_id'
 
+  before_validation :sanitize_postal_code
   before_save :format_cleanup
   before_create :chimp_subscribe
   before_update :chimp_check
@@ -254,6 +255,11 @@ class User < ApplicationRecord
 
     def default_country_to_usa
       self.country ||= "USA"
+    end
+
+    def sanitize_postal_code
+      return unless self.postal_code
+      self.postal_code = self.postal_code.delete(' ').upcase
     end
 
     def country_is_supported
--- a/app/validators/postal_code_validator.rb
+++ b/app/validators/postal_code_validator.rb
@@ -1,6 +1,6 @@
 class PostalCodeValidator < ActiveModel::Validator
-  ZIP_CODE_REGEX = /\A\d{5}(?:-\d{4})?\z/
-  POSTAL_CODE_REGEX = /\A[ABCEGHJKLMNPRSTVXY][0-9][ABCEGHJKLMNPRSTVWXYZ][0-9][ABCEGHJKLMNPRSTVWXYZ][0-9]\Z/
+  VALID_US_ZIP_CODE = /\A\d{5}(?:-\d{4})?\z/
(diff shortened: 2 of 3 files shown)


app/models/user.rb
```diff
@@ -256,6 +257,11 @@ def default_country_to_usa
       self.country ||= "USA"
     end
 
+    def sanitize_postal_code
+      return unless self.postal_code
+      self.postal_code = self.postal_code.delete(' ').upcase
+    end
```
jayroh: The setter `self.postal_code=` is 👍 but there's no need here for `self`s where they are referencing the `postal_code` getters.

app/models/user.rb
```diff
@@ -256,6 +257,11 @@ def default_country_to_usa
       self.country ||= "USA"
     end
 
+    def sanitize_postal_code
+      return unless self.postal_code
```
TheDudeWithTheThing: `return if postal_code.blank?`

no need for self
reads better

app/models/user.rb
```diff
@@ -256,6 +257,11 @@ def default_country_to_usa
...
+      self.postal_code = self.postal_code.delete(' ').upcase
```
TheDudeWithTheThing: no need for self here

I didn't know about `delete`, good stuff, I think always did `tr(' ', '')`

app/validators/postal_code_validator.rb
```diff
@@ -12,11 +12,9 @@ def validate(record)
       return
     end
 
-    postal_code = record.postal_code.delete(' ').upcase
-
-    if country.eql?(ISO3166::Country[:us]) && !postal_code.match(ZIP_CODE_REGEX)
+    if country.eql?(ISO3166::Country[:us]) && !record.postal_code.match(ZIP_CODE_REGEX)
       record.errors[:postal_code] << "should be 12345 or 12345-1234"
-    elsif country.eql?(ISO3166::Country[:ca]) && !postal_code.match(POSTAL_CODE_REGEX)
+    elsif country.eql?(ISO3166::Country[:ca]) && !record.postal_code.match(POSTAL_CODE_REGEX)
```
jayroh: What do you think about extracting these two conditionals out to more intention revealing private predicate methods?  Like `valid_us_zip_code?` and `valid_us_postal_code?` ? It's self documenting and reduces the cognitive friction a little bit more.

app/models/user.rb
jayroh: `return if postal_code.blank?` would avoid the unless.

app/validators/postal_code_validator.rb
TheDudeWithTheThing: careful now, `valid_CA_postal_code?`

jayroh: Oopsies

app/models/user.rb
tobymurray: Works for me - changed in  967e880

app/models/user.rb
tobymurray: second `self` removed in 967e880. I'm not super crazy about the `delete`, as I don't find it intuitive that it removes all matches, but at least it's a word?

app/validators/postal_code_validator.rb
tobymurray: Do you mean just the latter half of the conditional, or the whole conditional?

The conditional is doing two distinct things - checking what country the record is describing, and then checking the validity of the postal code in that country. I personally don't think `valid_us_zip_code` makes that more clear, and the extra layer of abstraction obscures the meat of the code from what is already what feels to me like a manageable context. Naming a function `valid_ca_postal_code` feels redundant when it's in a class called `PostalCodeValidator`.

Maybe a rename of the constants would provide a similar benefit? `VALID_CANADIAN_POSTAL_CODE` and `VALID_US_ZIP_CODE` or something?

app/models/user.rb
tobymurray: Done in 967e880

app/validators/postal_code_validator.rb
markottaviani: The notion that POSTAL_CODE_REGEX = Canada and ZIP_CODE_REGEX = USA isn't obvious at least for me as a non frequent snail mailer to Canada.    So I'm a fan of renaming the Regex constants like @tobymurray mentioned.

PR #662, opened 2017-10-29
437 add country to user form

tobymurray:
Incorporates PR #661, so hoping that gets merged shortly. This adds the country to the `user` form. I went for the easiest UI I could think of for the first pass.

image


markottaviani:
LGTM, will merge this in after #661 assuming @TheDudeWithTheThing  doesn't have any items of concern

Issue #590, opened 2017-07-13
Two Levels of Volunteers


markottaviani:
> The goal is to have two levels of volunteers and to not delete any, unless they are locked for business purposes.  

> We have many that sign up and don't do anything or much, but I would rather keep them on as supporters and hope they step up one day, or donate or pass on good words about us.  These would be Volunteer Supporters and not have access to DMS staff resources, just their profile.

> I would have a second group called Volunteer Administrators or something that are the active volunteers, that have positions, jobs, need DMS access.  If someone gets active from the supporter group I would move them to Administrative group.

> So, I would add all new people to just Volunteer Support, and once I see them being active move them up, and if they are not active move them back.

Feature Biz Contacts:
* Jen 
* Laurie
* Joanne

markottaviani:
Pending Clarification on the following:

If someone is considered "Volunteer Support" they can:
* login to DMS
* Access, edit and update their profile
* Do anything an anonymous public user can do
* Be automatically added to the volunteer mailing list


Should they be able to:
* View profiles of other users?
* View private dog data, or only see the standard public dog information?
* be granted any other roles or permissions (Adoption Coordinator, Reference Checker, Foster, etc.) or does their account need to be "Upgraded" First?

markottaviani:
@blai00  Going to review this with the biz folks tomorrow, hopefully you can get started soon.  LMK if you have any questions.  Thanks!

blai00:
hey mark, just read the requirements for this new change. just wondering when we see them "active" do they automatically become volunteer administrator or is it that someone who is already a volunteer administrator will move them up once they express their interest in helping out more to someone.

markottaviani:
Someone with admin rights would need to apply the permission/setting to the
Volunteer’s account.

Issue #411, opened 2015-09-20
Rebuild Authorizations


markottaviani:
Refactor into a proper permission system to better manage user permissions.

markottaviani:
https://github.com/elabs/pundit
https://github.com/CanCanCommunity/cancancan

TheDudeWithTheThing:
nice, thredded just moved to pundit so wanted an excuse to play with it

markottaviani:
Cool, let me know before you dive into this one.  It's probably worth talking to the Business and figuring out what the proper business roles are these days.

TheDudeWithTheThing:
#446 is a duplicate of this no?

markottaviani:
Nope, updated the description.  This is Authorization, the other is Authentication.

markottaviani:
Stuff to check out: 

rolify and cancancan — Rolify allows the application to assign roles such as User or Admin on a User object. Then CanCanCan allows you to take those roles and apply authorization policies.

PR #663, opened 2017-11-03
added volunteer support as a new user permission

blai00:
Connects to #590 

I need to figure out why it keeps creating that new db/schema. I have no idea why it keeps doing that.


markottaviani:
Search the code on your computer, you must have an active record migration
file that references that table.

markottaviani:
Reflecting on this some more, I think we should implement it a little differently.  

Instead of a new attribute called "Volunteer Support" let's just call it "Active".

If User.active==true (and their account isn't locked) that user should be able to:
* Do everything that a user can do today

However if User.active ==false   they should NOT be able to:

- [ ] View profiles of other users
- [ ] View private dog data, or only see the standard public dog information
- [ ] be granted any other roles or permissions (Adoption Coordinator, Reference Checker, Foster, etc.), account needs to be upgraded in some why to an "Active Volunteer" account

They should be able to:
- [ ] login to DMS
- [ ] Access, edit and update their profile
- [ ] Do anything an anonymous public user can do
- [ ] Be automatically added to the volunteer mailing list

Be sure to include rspec tests to verify these items 👍

blai00:
do we want to assume that a new volunteer is an active member (set default to true) or do want them to be set to false and allow an admin to change it?

markottaviani:
Good question, let’s default to false.

markottaviani:
Digging into this now.

spec/features/edit_user_profile_spec.rb
```diff
@@ -1,6 +1,6 @@
 feature 'Edit User Profile', js: true do
   let!(:admin) { create(:user, :admin) }
-  let!(:user) { create(:user) }
+  let!(:user) { create(:user, active: TRUE) }
```
tobymurray: `true` instead of `TRUE`?

spec/controllers/dogs_controller_spec.rb
```diff
@@ -105,9 +101,9 @@
 
       it 'Only adoptable, adoption pending or coming soon dogs should be displayed' do
         get_index
-
         expect(assigns(:dogs)).to match_array([adoptable_dog, adoption_pending_dog, coming_soon_dog])
       end
+
```
tobymurray: It looks like a newline was added in here - does rubocop complain about it?

db/migrate/20171103212000_add_active_users_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddActiveUsersToUsers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :active, :boolean, default: false, null: false
```
tobymurray: Maybe worth adding a `comment` to explain what `active` means?

app/models/user.rb
```diff
@@ -163,6 +166,8 @@ class User < ApplicationRecord
   scope :boarding_buddy,          -> (status = true) { where(boarding_buddies: status)}
   scope :social_media,            -> (status = true) { where(social_media_manager: status)}
   scope :graphic_designer,        -> (status = true) { where(graphic_design: status)}
+  scope :active_volunteer,        -> (status = true) { where(active: status)}
+  scope :not_active_volunteer,    -> (status = false) { where(active:status)}
```
tobymurray: Consider `inactive`?

app/models/user.rb
```diff
@@ -74,10 +74,13 @@
...
+#  boarding_buddies             :boolean          default(FALSE), not null
 #  medical_behavior_permission  :boolean          default(FALSE)
-#  boarding_buddies             :boolean          default(FALSE)
-#  social_media_manager         :boolean          default(FALSE)
-#  graphic_design               :boolean          default(FALSE)
+#  social_media_manager         :boolean          default(FALSE), not null
+#  graphic_design               :boolean          default(FALSE), not null
+#  country                      :string(3)        not null
+#  active                       :boolean          default(FALSE), not null
+#
```
tobymurray: Extra comment line

app/helpers/sessions_helper.rb
```diff
@@ -35,6 +35,10 @@ def signed_in?
     !current_user.nil?
   end
 
+  def active_user?
+    current_user.active? unless current_user.nil?
```
tobymurray: Maybe slightly easier to read with the "safe navigation" operator? `current_user&.active?

app/views/dogs/index.html.erb
```diff
@@ -10,7 +10,7 @@
 
       <% if session[:mgr_view] == true %>
         <%= link_to("Gallery View", {:controller => 'dogs', :action => 'switch_view'},:class => "btn ") %>
-      <% else %>
+      <% elsif current_user.active == true  %>
```
tobymurray: I think the `== true` is extraneous here.

spec/controllers/users_controller_spec.rb
```diff
@@ -139,11 +119,47 @@
...
 
-      it 'returns the translator team mebers' do
+      it 'returns the translator team members' do
         smith = create(:user, name: 'Jane Smithbot', translator: TRUE)
         get :index, params: { translator: TRUE }
         expect(assigns(:users)).to match_array([smith])
       end
+      
+      it 'returns the active team members' do
+        smith = create(:user, name: 'Jane Smithbot', active: TRUE)
```
tobymurray: Same comment as elsewhere `TRUE` -> `true`?

spec/controllers/users_controller_spec.rb
```diff
@@ -160,41 +176,85 @@
...
+      it 'is not able to modify user permissions' do
+        expect { change_permissions_request.call }.to_not change { test_user.reload.admin }
       end
+
+      it 'is able to update own profile' do
+        patch :update, params: { id: active_user.id, user: attributes_for(:user, city: 'New York') }
+        expect(active_user.reload.city).to eq 'New York'
+      end
+
+      it 'is not able to edit another user\'s profile' do
```
tobymurray: Use double quotes instead of escaping?

spec/controllers/users_controller_spec.rb
```diff
@@ -160,41 +176,85 @@
...
+        expect { change_name_request.call }.to_not change { test_user.reload.name }
+      end
+
+    end
+
+    context 'logged in as inactive active user' do
```
tobymurray: Looks like an extra `active` here? Otherwise it's not clear what it means.

spec/features/manage_dogs_spec.rb
```diff
@@ -2,22 +2,112 @@
...
+      expect(page).to have_content(test_dog.name.titleize)
+    end
   end
+
+  context 'Logged in as Inactive User' do
+    before(:each) do
+      sign_in(inactive_user)
+    end
+
+    scenario 'can not use the Manager View' do
```
tobymurray: `can not use` -> `cannot view`?

tobymurray:
Did you run `rubocop` to ensure you're not introducing new violations?

tobymurray:
Is it possible for an `admin` to be `active = false`? What does that mean? If not possible, maybe worth enforcing?

app/controllers/users_controller.rb
```diff
@@ -252,6 +255,15 @@ def init_fields
     @foster_users = User.where(locked: false).order("name")
   end
 
+  def active_user
+    redirect_to(root_path) unless current_user.active?
+  end
+
+  def show_other_users
```
tobymurray: With the first condition being `current_user?(@user)` it comes across as a bit confusing to me that this is called `show_other_users` - the same check is applied to view one's own profile. Perhaps something like `allowed_to_see_user` or similar? Naming is hard...

app/controllers/users_controller.rb
```diff
@@ -76,6 +76,8 @@
 class UsersController < ApplicationController
   before_action :authenticate
   before_action :correct_user, only: [:edit, :update]
+  before_action :active_user, only: [:index]
```
tobymurray: Purely style, but given that these only apply to a single action would it be easier to get rid of the `before_action` and the helper method and just call the logic directly?

app/views/layouts/_topbar.html.erb
```diff
@@ -66,15 +66,17 @@
                     <% if can_edit_my_adopters? || can_edit_all_adopters? %>
                       <li><%= link_to "Applications",  :controller => "adopters", :status => "active" %></li>
                     <% end %>
-                    <li><%= link_to "DNA List", :controller => "banned_adopters" %></li>
-                    <li><%= link_to "Manage Dogs", :controller => "dogs", :status => "active" %></li>
+                    <% if current_user.active? || current_user.admin? %>
```
tobymurray: Left this comment on the PR, but this might be redundant - are there any cases where an `admin` would not be `active`?

app/views/users/_form.html.erb
```diff
@@ -428,6 +428,10 @@
           <%= f.check_box :dl_locked_resources %>
           <span><strong>Download Restricted Resources</strong> - This person can access Restricted folders in the Staff Resources section.  Download Resources permission needed too.</span>
         </label>
+        <label class="checkbox">
+          <%= f.check_box :active %>
+          <span><strong>Active Volunteer</strong>- This person is an active volunteer. </span>
```
tobymurray: Shouldn't change anything, but extra whitespace at the end. Maybe worth a more descriptive comment - what makes someone active, or what makes an inactive volunteer different from an active volunteer?

app/models/user.rb
```diff
@@ -163,6 +166,8 @@ class User < ApplicationRecord
   scope :boarding_buddy,          -> (status = true) { where(boarding_buddies: status)}
   scope :social_media,            -> (status = true) { where(social_media_manager: status)}
   scope :graphic_designer,        -> (status = true) { where(graphic_design: status)}
+  scope :active_volunteer,        -> (status = true) { where(active: status)}
```
tobymurray: Maybe I don't know what's going on, but why use `status = true`? Could this be `where(active: true)`?

markottaviani:
Thanks for going through the code.  There is probably an opportunity to consolidate down the Admin and Active and Locked booleans into a single "Account Type".  As they're all mutually exclusive .

blai00:
Should I look into a way for that to happen? Or is that something that needs tobe discussed with OPH first to determine what they want and their ideal design?

markottaviani:
@blai00  That's probably going to be another big change, I'd kind of rather ship this first and then go back and refactor the logic.   Possibly as part of #411

blai00:
do you want me to go over Toby's changes, add all the extra comments and fix the small details and then we can ship it? Or as is is good for now?

markottaviani:
@blai00 sure, go ahead and work through Toby's suggestions.  Also run ```rubocop``` on each of the changed .rb files and correct any violations.  Only one I'm aware of that you can ignore for now is the "Class has too many lines" stuff.

blai00:
hey @markottaviani finished going through Toby's comments and put all .rb files in rubocop.

app/controllers/users_controller.rb
markottaviani: Possibly, but since all of these are related to user permissions it does keep all that logic in one place.  Hopefully it will make it easier to migrate us out to a halfway decent permissions solution later.

app/views/layouts/_topbar.html.erb
markottaviani: Not critical, but possible, was planning on addressing in a PR for refactoring Groups/Permissions.

app/controllers/users_controller.rb
TheDudeWithTheThing: yeah makes it easier to expand if we were to add `:delete` or `:give_free_puppy` or something down the line.

db/migrate/20171103212000_add_active_users_to_users.rb
TheDudeWithTheThing: you mean a DB comment?

tobymurray: Yes - only really relevant for development, but a QOL feature I've recently found valuable.

markottaviani: Ahhh https://blog.bigbinary.com/2016/06/21/rails-5-supports-adding-comments-migrations.html

app/models/user.rb
markottaviani: Probably, I just haven't figured out the right way to do it.  Stripping it down to just that throws errors about having too many arguments.

markottaviani:
Great work on this @blai00 I think this might be ready to go.  Will push this out tomorrow unless there are any remaining concerns

Issue #595, opened 2017-07-15
Adoption Coordinators should be able to edit Applicants References


markottaviani:
Source:  https://www.facebook.com/groups/348416295253712/permalink/1329164933845505/

As an Adoption Coordinator I should have the ability to update an Applicant's references, as these frequently change.   The changes should be automatically logged in the comments section.

blai00:
should there be a new permission for adoption coordinators that could edit it?
should this also be available for admins or just specifically for adoption coordinators

markottaviani:
No new special permissions.  Same folks that can edit an adoption applicant's address should be able to update the references.  Ideally try to follow that same behavior.

PR #669, opened 2017-11-11
WIP #595 Adoption Coordinators should be able to edit references

taratip:
Reference part is now editable for adoption coordinators. Will work on logging changes automatically in comment section.

Connects to #595


app/controllers/adopters_controller.rb
```diff
@@ -176,6 +176,7 @@ def adopter_params
                                     ],
                                     references_attributes:
                                     [
+                                      :id,
```
markottaviani: Since you're not modifying the id value of the reference, I don't think this line is required.

taratip: I tried without id, but it inserted instead of updating.

app/models/adopter.rb
```diff
@@ -39,6 +39,7 @@
 #  completed_date      :date
 #  county              :string
 #
+require 'pry'
```
markottaviani: This should get removed, leftover from debuggin?

config/database.yml.tara
```diff
@@ -0,0 +1,7 @@
+development:
+  adapter: postgresql
+  database: RescueRails_development
+
+test:
+  adapter: postgresql
+  database: RescueRails_test
```
markottaviani: This file shouldn't get committed since it's personal.

spec/features/apply_to_adopt_spec.rb
```diff
@@ -302,19 +302,19 @@
...
+
+    expect(page).to have_field("adopter_references_attributes_1_name", with: "Second Reference")
+    expect(page).to have_field("adopter_references_attributes_1_email", with: "second@reference.org")
+    expect(page).to have_field("adopter_references_attributes_1_phone", with: "(222) 222-2222")
+    expect(page).to have_field("adopter_references_attributes_1_whentocall", with: "After 2pm")
+
+    expect(page).to have_field("adopter_references_attributes_2_name", with: "Third Reference")
+    expect(page).to have_field("adopter_references_attributes_2_email", with: "third@reference.org")
+    expect(page).to have_field("adopter_references_attributes_2_phone", with: "(333) 333-3333")
+    expect(page).to have_field("adopter_references_attributes_2_whentocall", with: "After 3pm")
```
markottaviani: Nice improvement 👍

app/views/references/_reference.html.erb
```diff
@@ -1,17 +1,24 @@
...
+      </td>
+      <td>
+        <%= rf.text_field :whentocall %>
+      </td>
+      <td>
+        <%= rf.submit "Save", class: "btn btn-primary" %>
+      </td>
+    </tr>
+  <% end %>
+<% end %>
```
markottaviani: From taking a look at how this appears on the site, it winds up looking really really wide.  Given all the small screens people are using  it might be worth considering a different layout, maybe something a little more vertical?

markottaviani:
Added some comments inline

config/database.yml.tara
taratip: I'll remove the file

taratip:
I changed to a layout with vertical headings for reference instead. Let me know if it looks good.

markottaviani:
Looking good! Two adjustments:

1) Can you make the edit interface behavior (i.e. toggling of editable form fields) the same as the *Edit Info* / *Save Info* box directly to the left of the references.  Instead of the word "info" you can use the word "references" on the buttons.  

2) Since all references fields are part of the same form, just have a single submit button for all three references.  Otherwise the existing setup gives the impression that you need to edit and save each reference individually.  

Thanks!

app/helpers/current_helper.rb
```diff
@@ -0,0 +1,3 @@
+module CurrentHelper
+  thread_mattr_accessor :current_scope_user
```
tobymurray: This looks a bit arcane to me - you're creating a thread-global constant? Without being familiar with the code, `CurrentHelper` isn't very intuitive. What problem are you solving with this?

markottaviani: It looks like a lot of this was written to try and get the name of the current user that was logged in so that the modification can be logged to the system.  However I think ```Auditable``` takes care of this.  @TheDudeWithTheThing I know you wrote a lot of the audit logic.  My spider sense things that some of the stuff in this PR shouldn't be needed if you can take a look too.

taratip: I tried to get the name of the current user in Reference model so that the changes can be logged in comment section.

taratip:
I made some adjustments based on feedback. Please let me know if it works as expected. Regarding the thread-global constant, is it alright to implement this way? If not, can you suggest if there's another way I can get the current user in Reference model.

markottaviani:
Thanks @taratip !  I'm not sure off the top of my head the preferred implementation for getting the current user, so I'm going to merge this into a new feature branch in our repo so I can more easily mess around with it and possibly push a few changes to it.

Issue #580, opened 2017-06-10
Security - Login Enhancements


markottaviani:
Security functionality need some love

Consider switching to Devise or another gem to standardize and improve instead of using the existing setup. 

Try to deal with
* Brute Force Passwords
* Enable CAPTCHA when multiple logon attempts
* Track IP and Logins
* Email Notifications on Logon
* Address MutliFactor

markottaviani:
@TheDudeWithTheThing if you'd like to try and tackle something a little more interesting, this might be up your ticket.

tobymurray:
@markottaviani - are there any ways an unauthenticated user (anonymous) is different from a user who has signed up but not set their password/signed in? I guess someone who has signed up has the opportunity to go through the "forgot password" workflow - anything else you can think of?

Is there any user registration (volunteer or staff) that is not driven by an administrator? E.g. a "sign up" button somewhere?

markottaviani:
I think that’s all covered.  Only an admin can add a new user.

tobymurray:
What does it mean for an account to be locked or unlocked? I see that they're unsubscribed from mailings, and it looks like they can't log in, but I'm not seeing why they would be locked/unlocked. What kind of things would cause an account to be locked, and what kind of things would cause an account to be unlocked?

markottaviani:
Locked is situation where someone totally leaves the rescue.  Maybe they’re
“fired” or totally quit.  We don’t delete the accounts, as it makes it
easier to be aware if someone tries to revolunteer 5 year’s later or if
someone has comments on the system their name is retained.

So we do wind up with mutually exclusive types of user accounts:


ADMINS - Can add new users and modify anyone’s permissions

ACTIVE - Active Staff/volunteers that are contributing to the rescue.  Gain
access to lots of private information and ability to be granted additional
roles and responsibilities.


INACTIVE - people that said they wanted to volunteer in some way , but
either don’t do anything or haven’t done much of anything in a while.
Their access should be no different that a public anonymous user on the
website other than they have a Profile they can edit, and they should get
subscribed to the mailing list.  The hope is that they might step up in the
future and be more active.  so we might email them if there is an event in
their area and we need volunteers, etc.


LOCKED - They have either totally quit/vanishes or been terminated from any
involvement with the rescue.  We don’t delete user accounts, this provides
a record of anyone that left and why.  Also keeps names tied to
commits/audit trails in the system.


[...]

PR #673, opened 2017-11-15
#580: Add not-null constraints to email and encrypted password

tobymurray:
Exactly what it sounds like, email and password are used to sign in so they can't be `null`. Enforce that at the database by adding the `NOT NULL` constraint. I imagine data that does not meet these requirements could be found with e.g.

`SELECT id, name, email, encrypted_password FROM users WHERE email IS NULL OR encrypted_password IS NULL;`

I would hope that nothing is returned by that query, if there is then we'll need to figure out why.

### DIFF
--- a/db/migrate/20171115013311_add_not_null_to_user_email_and_password.rb
+++ b/db/migrate/20171115013311_add_not_null_to_user_email_and_password.rb
@@ -1,0 +1,6 @@
+class AddNotNullToUserEmailAndPassword < ActiveRecord::Migration[5.0]
+  def change
+    change_column_null(:users, :email, false)
+    change_column_null(:users, :encrypted_password, false)
+  end
+end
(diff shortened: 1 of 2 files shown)


markottaviani:
So, there are no accounts with null email addresses.   However there are a lot of accounts with null encrypted passwords.   

When an account it created for a new staff member an default password is not set for them (too risky for them to forget to change it).  Instead the password stays as null and they need to use the Forgot Password feature to create their initial password.   

However there are many instances where an account is created for a volunteer and they never log in, possibly because whatever they do doesn't require them to really use the site to get their work done.  Or they just ghosted.

tobymurray:
Thanks for the thorough explanation. I'm looking at the simplest way to cut over to Devise, and this kind of stuff is definitely necessary to know about.

markottaviani:
For sure! Feel free to ask any questions that pop up.

PR #681, opened 2017-11-26
Adoption Coordinators Should be Able to Edit References

markottaviani:
connects to #595


app/models/reference.rb
```diff
@@ -28,5 +28,40 @@
 #
 
 class Reference < ApplicationRecord
+  include Auditable
+
   belongs_to :adopter, class_name: 'Adopter'
+
+  after_update :audit_changes
+
```
markottaviani: I think the ```Auditable``` concern will work right (and we can get rid of a lot of this code) by just adding something kind of like ```attr_accessor :updated_by_admin_user```.  However it appears that the reference model isn't able to pick up the attribute when it gets passed in at line 102 of the adopters_controller.  Google-Fu is failing me so far.  

@TheDudeWithTheThing @tobymurray

taratip: I think it still needs to store current user... I found the request_store gem. It acts like Thread.current[], but clears itself after each request.

markottaviani: The Adopter model is able to take care of this because the current user is [passed in during the update call](https://github.com/ophrescue/RescueRails/blob/master/app/controllers/adopters_controller.rb#L102).   I haven't been able to figure out how to get that value available during the call of the Audit Concerns, like it is already w/ the adopter model.

markottaviani:
Bounced a few ideas around with @TheDudeWithTheThing .   Auditable concern winds up turning into a very ugly thing to deal with when we get into nested attributes any way we slice it.  Requiring weird code all over the place.  

There is a lot of interest LOTS auditing all over the place from the Rescue, so we're think we might want to explore transiting the existing audits to use https://github.com/airblade/paper_trail and probably improve the existing audits by put them somewhere other than the existing comment stream.    

Going to park this PR for right now until we can figure out what that might need to involve.

spec/rails_helper.rb
```diff
@@ -12,14 +12,14 @@
...
+Capybara.register_driver :selenium do |app|
   capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
-    chromeOptions: { args: %w[headless disable-gpu window-size=1366,768] }
+    chromeOptions: { args: %w[headless disable-gpu window-size=1366,2000] }
   )
   Capybara::Selenium::Driver.new app, browser: :chrome, desired_capabilities: capabilities
 end
 
-Capybara.javascript_driver = :headless_chrome
+Capybara.javascript_driver = :selenium
```
tobymurray: Out of curiosity, what was failing here? The commit message said "PNGs"?

app/assets/javascripts/adoption_form.js
```diff
@@ -66,5 +66,30 @@ $( function() {
         select: itemSelected,
         source: remoteSource
       });
+
+    var isRefEditing = false;
+
+    $('#references').on('click', 'a#toggle-edit-ref', function() {
+      if (isRefEditing) {
+        // find all disabled fields in .edit-reference and enable them
+        $('form.edit-reference .to-disable').attr('disabled', 'disabled');
```
tobymurray: Not my forte, and is probably readily apparent during use, but is this right: `.attr('disabled', 'disabled');`? It looks a bit suspicious. Two things:
- the comment says it's enabling them but it looks like it's setting it `disabled`,
- in XHTML `disabled` an attribute, but in HTML5 it's a property - should this be a property here? I imagine functionally equivalent, but not sure what the browser support story is. E.g. `.prop( "disabled", true );`

spec/rails_helper.rb
tobymurray: Looks like this is addressing https://github.com/mattheworiordan/capybara-screenshot/issues/211 ?

app/assets/javascripts/adoption_form.js
taratip: @tobymurray 
- The comment is wrong.
- You are right, it should be .prop() according to jQuery1.6+ documentation. 
@markottaviani 
Do you want me to work through Toby's suggestions?

markottaviani:
Go for it, thanks!

TheDudeWithTheThing:
@taratip @markottaviani what's going on here?  needs review?  we redoing it?

taratip:
@TheDudeWithTheThing I think @markottaviani wanted to explore a different audit instead of using Auditable concern. However, I'm not sure where we are right now.

markottaviani:
Yup, this is parked until #704 gets done so we can do good auditing.

markottaviani:
Okay, I think this is pretty close to being ready to ship! 
https://cl.ly/3A0u230W0c1m

Only issue I really see right now is with the Field Name.  If someone changes the name of a reference the audit report just says "Name" for the field.  Can be kind of confusing since it isn't clear if we're talking about the name of the adopter or a reference or what.    @TheDudeWithTheThing any thoughts on this?

https://cl.ly/3S3q12354630

PR #697, opened 2017-12-05
Modifies development to use file storage instead of S3

### LINKED ISSUE #648: Fix Paperclip to use local file storage in development
Right now local dev tries to use S3 for file uploads. We can make it so development uses local storage instead so you don't need an S3 account for dev work.

TheDudeWithTheThing:
Fixes #648

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -54,15 +54,6 @@ class Attachment < ApplicationRecord
   AGREEMENT_TYPE_CODE_OF_CONDUCT = 'code of conduct'
 
   def download_url(style_name = :original)
-    s3 = Aws::S3::Client.new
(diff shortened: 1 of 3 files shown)


app/models/attachment.rb
```diff
@@ -54,15 +54,6 @@ class Attachment < ApplicationRecord
...
-    resource = Aws::S3::Resource.new(client: s3)
-
-    @bucket ||= resource.bucket(ENV['S3_BUCKET_NAME'])
-
-    @bucket.object(attachment.s3_object(style_name).key).presigned_url(
-      :get,
-      secure: true,
-      expires_in: 300,
-      response_content_disposition: "attachment; filename=#{attachment_file_name}")
+    attachment.expiring_url(3600, style_name)
```
markottaviani: Nice!

Issue #671, opened 2017-11-13
Dogs Index Render Bug


blai00:
Steps to Reproduce
- Log in as admin
- Go to /dogs (dogs index)
- Select Manager View ("Dogs Manager" should be the title of the page)
- Log out 
- Go to /dogs (dogs index)

At this point, you should still see the title "Dogs Manager" rather than "Our Dogs"

markottaviani:
Bug Confirmed. 

Desired behavior after going through the steps that @blai00 mentions above is that the header should say "Our Dogs" when looking at the Gallery, however it shows "Dogs Manager" when looking at the gallery even when logged out.

blai00:
Since I found the bug, I will investigate for the time being of what solution can work for it.

PR #698, opened 2017-12-05
671 dogs index render bug

blai00:
connects to #671

### DIFF
--- a/app/views/dogs/index.html.erb
+++ b/app/views/dogs/index.html.erb
@@ -20,15 +20,12 @@
 <%= will_paginate @dogs %>
 
   <% if signed_in? && session[:mgr_view] == true %>
-
     <%= render 'manager' %>
-
+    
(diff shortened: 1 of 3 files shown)


app/views/dogs/index.html.erb
```diff
@@ -20,15 +20,14 @@
 <%= will_paginate @dogs %>
 
   <% if signed_in? && session[:mgr_view] == true %>
-
+    
     <%= render 'manager' %>
-
+
```
markottaviani: @blai00 what are you using for an editor/IDE?

blai00: brackets for MAC, should I be using something else? I am not particularly fond of it so any suggestion would be nice

markottaviani: Try out https://code.visualstudio.com/

Issue #684, opened 2017-12-01
Make Dog New and Edit Pages the same


markottaviani:
There are a bunch of Dog fields that can only be set after the dog is created via the edit page.  
* Source Shelter
* Photos
* etc

Update the Dog new page so that these fields can be set there.

PR #699, opened 2017-12-05
Made Dog New and Edit Pages the same

taratip:
connects to #684


markottaviani:
Do you think this could be consolidated to a single file?  Similar to how the user new and edit views are taken care of?

app/views/dogs/new.html.erb
```diff
@@ -78,6 +84,18 @@
...
 			</div>
 		</div>
+		<div class="control-group">
+			<%= f.label :adoption_date, "Adoption Date", :class => 'control-label'  %>
+			<div class="controls">
+				<%= f.text_field :adoption_date %>
+			</div>
+		</div>
+		<div class="control-group">
+      <%= f.label :sponsored_by, "Sponsored by", :class => 'control-label' %>
```
tobymurray: Use hashrocket style? `:class => 'control-label'` -> `class: 'control-label'`

taratip:
I think it could be consolidated to a single file. I'll change it.

Issue #682, opened 2017-11-29
Migrate from Auditable Concern to more flexible Auditing System


markottaviani:
The Rescue <3 Auditing stuff, we've hit a point where we've probably outgown `Auditable`.  There is a LOT of stuff that they'd love to be able to audit.  

Objectives
* Transition to either the  `paper_trail` or `audited` gem 
* Log Activity to "Somewhere else" other than the comment section, maybe a History Tab/ History Table. 
* See if we can come up with a way to migrate the existing history in Comments into the new format/table. 
* Transition all existing Auditable activities.

markottaviani:
Started working with the `audited` gem, the basic implementation was pretty easy.  Trick is going to be displaying and working with the audited_changes.

By default the changes are stored in a YAML format in the database
```
---
status:
- adoptable
- adopted
adoption_date:
- 
- 2017-12-09
```

Calling `<%= audit.audited_changes %>` will spit out something like

```
{"status"=>["adopted", "return pending"], "adoption_date"=>[Sat, 09 Dec 2017, nil], "coordinator_id"=>[nil, 2]}
``` 
And I haven't had much luck so far parsing that into anything super useful.  

`audited` supports Postgresql's native json and jsonb objects and rails seems to have some [nice support for them as well](https://nandovieira.com/using-postgresql-and-jsonb-with-ruby-on-rails).   

Since we're only running Postgresql 9.3.x in production, I think we need to do some upgrades to at least 9.4 for some capabilities.  Luckily we're running Amazon RDS so the upgrades should be pretty painless.

PR #704, opened 2017-12-25
682 audit capture and display

markottaviani:
Connected to #682 

TODO
- [x] Tests
- [x] Add Basic Auditing for Dogs and Adoption Applications
- [x] Add History view to Adopter
- [x] Track nested changes
- [ ] Provide an audit entry when a file attachment is added or overwritten.  

Possibly split into separate items:
- [x] Show a name value instead of an ID when a foreign key is changed.
- [x] See if there is a way to blend History and Comments into one view (All)


markottaviani:
@TheDudeWithTheThing  Trying to figure out how to pull off the "All" tab: 
 https://cl.ly/3K0B2Z0W0f3M 

How to blend the Comments sorted by time and the new audits sorted by time.

markottaviani:
Possible option for dealing with foreign keys https://johnromani90.github.io/2017/auditing-rails-models/

markottaviani:
@ophrescue/site-developers anyone want to try getting the foreign key stuff working?  Trying to achieve what's mentioned in that blog post above, but I'm not getting anywhere with it.

TheDudeWithTheThing:
@markottaviani doesn't look too bad, I'm not sure I like how this guy is doing it but will have a look.  Example of something that is coming up with an ID instead of the name?

markottaviani:
Yeah, totally fine with a different implementation.  

A few examples from the Dog model:
* primary_breed_id
* coordinator_id
* shelter_id

For Adopter model:
* assigned_to_user_id

app/models/adopter.rb
```diff
@@ -161,6 +163,10 @@ def chimp_subscribe
     self.is_subscribed = true
   end
 
+  def comments_and_audits
+    comments + audits
```
markottaviani: Awesome

TheDudeWithTheThing: but did it work??

markottaviani: It does!  Adding sorting now.

app/models/adopter.rb
```diff
@@ -118,6 +120,10 @@ def attributes_to_audit
     %w[status assigned_to_user_id email phone address1 address2 city state zip]
   end
 
+  def audits_and_associated_audits
+    (audits + associated_audits).sort_by(&:created_at)
```
markottaviani: @TheDudeWithTheThing any suggestions on how to sort this stuff?

TheDudeWithTheThing: This is what I tried too, it works on console but not in the app right? So strange, I have a theory

markottaviani: Think I got it working, change pushed

app/models/adopter.rb
```diff
@@ -161,6 +167,11 @@ def chimp_subscribe
     self.is_subscribed = true
   end
 
+  def comments_and_audits_and_associated_audits
+    clean_comments = comments.to_a.delete_if { |obj| obj.id == nil }
```
markottaviani: There is an extra comment with a nil id in here because of the comment.build earlier up in the view.  This is ugly and will need some cleaning up.

markottaviani:
Alrighty, I think we might have something shippable!  Thanks @TheDudeWithTheThing for the Foreign Key stuff!   

I'm going to push this over to staging tomorrow for folks to start poking at.

PR #711, opened 2018-01-30
bundle update 

markottaviani:
fix paperclip CVE
rails 5.0.6
pg set to not use 1.0 yet as it throws lots of errors.


markottaviani:
Now why are these travis tests failing and/or taking really long to run? I killed the last test run after letting it run for 45 mins.  Suite runs fine locally.

markottaviani:
Back in action!

PR #726, opened 2018-03-06
redirect tracking for  5k registration

markottaviani:
The 5k promotion team is starting a social media campaign for the upcoming OPH 5k.  They'd like to be able to track visitors to the `/5k` link via Google Analytics.  

This change creates an actual page which should provide some time for Google Analytics to fire off and collect the event and related details.    Suggestions on if there is a better way to pull this off are appreciated.


TheDudeWithTheThing:
Haha, I know nothing about tracking. I'd imagine you'd be able to do something in the URL with a GET param if you really wanted

markottaviani:
I think they just want to see the basic referrers, from social media, so I guess I'll take this for a test drive.

markottaviani:
Seems to work okay!

PR #746, opened 2018-04-19
Training Partner Page

markottaviani:
New webpage listed out training partners.


lazylester:
@markottaviani is there a reason the partners data is hard-coded in the views, vs stored in the db?

markottaviani:
This is expected to be mostly static content.

PR #750, opened 2018-04-22
Rails 5.1 and URL Login Redirect

markottaviani:
New Pull Request for Rails 5.1 and URL Login Redirect fix.


markottaviani:
@lazylester I caught your rails5.1 branch up to the master branch, and fixed the old ActiveRecord migrations so that travisci could run.  Looks like there are a few failing tests you might want to check out.  Feel free to pull/push with this branch.

lazylester:
@markottaviani I've pulled the feature/rails-51 branch, and it passes all tests (run using 'rake'). I see the travis-ci failures, but they pass on my machine. Any suggestions how to reproduce the travis failures?

lazylester:
the 'rake' command (with no arguments) runs all the spec files that are found by this pattern:
spec/\*\*\{,/\*/\*\*\}/\*_spec.rb

so it seems to pick up all the rspec specs ("246 examples 0 failures")

btw does travis-ci really run all the migrations to re-create the database? this seems fraught with possible problems. Isn't it possible to load from db/schema.rb?

markottaviani:
Honestly not sure.  I’m able to replicate the test  failures on an Ubuntu
VM via `rspec` so I’ll look into them this week.

lazylester:
OK, I can see the failures now. I'll look into them. please stand by.

lazylester:
@markottaviani I have fixed the probs I can reproduce (they related to rspec itself), but there are two related to ActionMailer::Base.deliveries count that pass for me but fail on Travis-ci. Can you let me know if they fail on your machine, please? Are you aware of any difficulties simulating ActionMailer deliveries on Travis-ci?

lazylester:
OK, I don't understand, but now the build is passing. I'm a travis-ci noob, so I don't really understand how this could happen. But it would seem to be good news. When I first pushed the changes, the build failed.

Gemfile
```diff
@@ -64,6 +62,7 @@ group :development, :test do
   gem 'pry-stack_explorer'
   gem 'rspec-collection_matchers'
   gem 'rspec-rails'
+  gem 'puma'
```
TheDudeWithTheThing: this seems weird, we should just switch everything to puma I'd think not just dev, want to do that in a separate PR?  @markottaviani are there any gotchas with our host for puma, it does seem to be the new hotness.

app/controllers/sessions_controller.rb
```diff
@@ -26,7 +31,8 @@ def create
     else
       sign_in user
       user.update_attribute(:lastlogin, Time.now)
-      redirect_to root_path
+      redirect_back_or(root_path)
+      #redirect_to root_path
```
TheDudeWithTheThing: any reason to keep this commented out vs. deleting it?

lazylester: no reason, other than preservation of the endangered cruft species :) 

I'll delete it

config/environments/production.rb
```diff
@@ -56,6 +56,7 @@
   # config.active_job.queue_adapter     = :resque
   # config.active_job.queue_name_prefix = "rescue_rails_#{Rails.env}"
   config.action_mailer.perform_caching = false
+  config.action_mailer.delivery_method = :smtp
```
TheDudeWithTheThing: Don't all our emails go through mailchimp?

TheDudeWithTheThing: ah I see, from the mailhopper config

👍

lazylester: I don't know, Shaun. I don't know how mailchimp works. If someone has some history with the mailer I'd appreciate some guidance.

TheDudeWithTheThing: I think you're good here, let's just make sure non mailchimp mails work still when we're on staging.

@markottaviani any idea which emails those are?

lazylester: how do you do staging for this app? Is there a staging server with public access somewhere?

Looks like mailchimp is used for User (un)subscribe and Adopter (un)subscribe (in app/jobs). Both the Adopter and User models have callbacks on create and update to trigger mailchimp emails.

lazylester: I note that the Travis-CI build is now failing again. It's the same timing issue related to sending emails that I just fixed on another  PR. I'll port the same fix into this PR. Please stand by...

lazylester: Build is passsing. Is there anything I need to do related to staging?

markottaviani: Everything in `app/mailers`, "transactional emails" goes out via smtp & sendgrid.com .

RescueRails only concern with Mailchimp is adding, removing and updated the mailing lists in Mailchimp. 

I'll make sure the email plumbing still works the same in the staging env.

Gemfile
markottaviani: I don't know, I only added this because Capybara was complaining about about puma missing when I was running the rspec suite.    Specific error was:

```
Capybara is unable to load `puma` for its server, please add `puma` to your project or specify a different server via something like `Capybara.server = :webrick`
```

markottaviani:
boom :shipit: nice work @lazylester

PR #751, opened 2018-04-23
Feature/rails 52

lazylester:
@markottaviani I need some help to understand the preferred workflow here. But hopefully you can merge this PR for rails 5.2 upgrade.


lazylester:
@markottaviani again the travis-ci failures on this PR relate to ActionMailer::Base.deliveries assertions in rspec. Can you reproduce this? The specs pass on my dev machine. I'm at a bit of a loss here.

In my previous (Rails 5.1) upgrade, the same thing happened, and then magically travis-ci became happy, with no intervention on my part!

lazylester:
OK nailed it... it's a timing issue related to ActionMailer#deliver_later. So I'm mocking it out. The delivery is tested in the mailer spec anyway, so there's no loss of coverage.

app/models/dog_searcher.rb
```diff
@@ -31,25 +33,24 @@ def initialize(params: {}, manager: false)
   end
 
   def search
-    @dogs = Dog.includes(:photos, :foster)
-    if @manager
-      @dogs = @dogs.includes(:adopters, :comments)
+    if @manager # manager view
+      @dogs = Dog.includes(:adoptions, :adopters, :comments)
```
TheDudeWithTheThing: were `:photos` and `:foster` not doing anything?

TheDudeWithTheThing: same with removing `with_includes`, so `:primary_breed` and removing photos from there is ok since it was a dupe, oops

TheDudeWithTheThing:
btw you can change the branch you're comparing to, looks like a lot of these changes are in the 5.1 branch

Hit edit for the title of the PR and swap the branch, or we can just review once 5.1 is merged.

lazylester:
Didn't understand "edit the title" comment. But anyway, yes this was built on Rails 5.1, so maybe it'll make more sense when that is merged.

(I'm still trying to figure out the appropriate workflow here, I've always been a sole contributor in the past.)

.travis.yml
```diff
@@ -21,7 +21,7 @@ env:
     - secure: "AnnnXIFPy93JvGO+DuqSpYGPd9MZNSxwciYKTcCV6Yx5htJwZFU6adAak9UWvXr8vjtfwrHJD/VqViC6GII7toVCF9O9XEc6ooHVHsLGxfqvl0uOy+prgZS0F8bYW8n6EjFrWPfywdEoMn8PsZ/9iNxPFC2tH8vVKmJAq/FnxUo="
 
 script:
-  - RAILS_ENV=test bundle exec rake db:migrate --trace
+  - RAILS_ENV=test bundle exec rake db:schema:load
```
markottaviani: @TheDudeWithTheThing what are your thoughts here?  

While I'm sure `db:schema:load` would be the faster way to run this,  the db:migrate runs pretty fast in travis already.   When we run a production deployment today, it's the migrations on the existing database that are running, not the db:schema:load.  So I care more about the migrations being always good than the full schema file.

TheDudeWithTheThing:
no worries, will wait on 5.1 merge then

but just so you know, at the top of this PR you can hit the edit button to edit the title of the PR, when you do that you can also change the base branch to compare this one to

[](!https://cl.ly/3P2i03192e3k)

.travis.yml
TheDudeWithTheThing: hmm, not bad I guess, but you need to run the migration locally first to get it into the schema file. At work we started loading just the schema since we have 4 years of migrations, if we can get away with running the migrations still I'm all for it.

app/models/dog_searcher.rb
lazylester: FYI this class is getting a major refactoring when the filter-by-breed feature is added. I stopped cleaning up here, for Rails 5.2 and deferred it to the next feature I'm working on (filter-by-breed).

.travis.yml
lazylester: It's not really a speed thing, but as we just found out, you can run into trouble running migrations and avoid all problems loading schema.rb. It seems to make little sense (to me) running through the entire history of the migrations when you already have a snapshot of the outcome.

lazylester: when deploying in production, you're running the most recent migration(s) with the current version of Rails (low risk). But when running the entire migration history, you're running all migrations (maybe from years ago) with the current version of Rails (could be problematic).

markottaviani: I'll take a CI problem any day of the week if it mitigates a production database migration problem.  Let's switch this back.

lazylester: happy to switch back to migrations iso schema load for CI, if you wish. However this does not increase the likelihood of catching production migration issues in CI (which would be indisputably good). 

A production migration problem (incremental migration) will be discovered in development. The "from-zero" migrations in CI simply introduce CI migration problems that do not pertain to development or production.

markottaviani:
Looks good to me, pushed this to our staging server, will probably ship this in the next 24-48 hrs

PR #765, opened 2018-04-30
Bootstrap 4.1 upgrade for home page

lazylester:
This leaves the build "broken", as it was when I pulled the baseline for this PR. There are also some design anomalies remaining from the baseline. The main issues resolved after including Bootstrap 4.1 were the topbar design, the carousel design, and mobile optimization of the "take action" section.

Since the topbar (= navbar) is included on all the pages, there are two versions, one for BS4.1 and one for the older version of BS.

I'm hoping that, as the design of this page is refined, going forwards, it can be done with this bootstrap 4.1 upgrade.


markottaviani:
Looks like all the BS 2.3.2 pages are chopped off on the left hand side:

https://cl.ly/0T3m21262k0B

PR #773, opened 2018-05-16
returned undocumented feature of sort by tracking_id

markottaviani:
Historically the dog gallery has always been sorted by dog tracking_id.  Adds back in and includes a test


lazylester:
@markottaviani  I'm curious about the use of tracking_id. Noting that it is intended here as a sort field, but it's not displayed in the view. So the perceived ordering seems to be random. Since the tracking_id is sequential, is the intention to display "least recent first"?

PR #774, opened 2018-05-16
Issue771 fix

lazylester:
cleans up the horribly overloaded DogsController#index method by splitting off the authenticated cases into their own #manager_index method, including its own manager_index view. This turns out to simplify many things.

### DIFF
--- a/app/controllers/dogs_controller.rb
+++ b/app/controllers/dogs_controller.rb
@@ -80,15 +80,6 @@ class DogsController < ApplicationController
     @dogs = case
             when params[:autocomplete] # it's autocomplete of dog names on the adopters/:id page
               Dog.autocomplete_name(params[:search])
-            when params[:commit] == 'Search' # search button was clicked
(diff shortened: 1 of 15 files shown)


app/controllers/dogs_controller.rb
```diff
@@ -101,6 +92,24 @@ def index
     end
   end
 
+  def manager_index
```
TheDudeWithTheThing: I forget the exact permissions here but this route should be protected somehow. `admin?` is there a dog manager role?  I really gotta get on those perms

TheDudeWithTheThing:
This is great, such an "oh yeah duh" type change, isolate all the manager specific things to their own route / controller action / views.

Good stuff!

app/controllers/dogs_controller.rb
TheDudeWithTheThing: ha, I asked you first!

It looks like we previously just looked for `signed_in?` so I think we're good

PR #777, opened 2018-05-18
Uses clearance gem for authentication

TheDudeWithTheThing:
https://github.com/ophrescue/RescueRails/issues/580

- [ ] This PR: Replace current auth with clearance gem
- [ ] Switch to Bcrypt password strategy
- [ ] Sign in guards for logging IP, failures, etc...
- [ ] Clean Up remaining code from original implementation


app/views/adopters/_adopter.html.erb
```diff
@@ -1,6 +1,6 @@
 <tr>
   <td><%= adopter.id %></td>
-  <td><%= adopter.adoption_app.ready_to_adopt_dt.blank? ? " " : adopter.adoption_app.ready_to_adopt_dt.strftime('%m/%d/%y') %></td>
+  <td><%= adopter&.adoption_app&.ready_to_adopt_dt.blank? ? " " : adopter.adoption_app.ready_to_adopt_dt.strftime('%m/%d/%y') %></td>
```
TheDudeWithTheThing: unrelated to this, I just have bad data locally and wanted to work around it

lib/password_strategies/woof_bark.rb
```diff
@@ -0,0 +1,28 @@
+module PasswordStrategies
```
TheDudeWithTheThing: this is our home rolled password strategy, once this is out we can force a forgot password and swap it to the default bcrypt version

markottaviani: Will do!

app/controllers/photos_controller.rb
```diff
@@ -29,6 +29,8 @@
 #
 
 class PhotosController < ApplicationController
+  before_action :require_login
```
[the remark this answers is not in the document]
markottaviani: nice catch!

markottaviani:
@TheDudeWithTheThing so will this require everyone to change their password, or is that in a later "phase 2" ?

TheDudeWithTheThing:
@markottaviani later phase 2, I'm trying to introduce the least amount of change here, which as you can see in the check list is just "use this gem".

The current password strategy is in a clearance compatible strategy I called "WoofBark", which is in this PR and allows people to log in how they do today.

TheDudeWithTheThing:
@markottaviani review for sure, can we smack it around on staging?

markottaviani:
So, still a few failing tests that should get looked at.   Also I tried to create a user account on the staging box without specifying a password and it bombed out.

```
users#create (NoMethodError) "undefined method `password_confirmation=' for #
```

TheDudeWithTheThing:
oh yeah I didn't touch the create user for some reason, do we not have tests for that?

markottaviani:
Also it looks like no matter what email address people enter (even one for a user that doesn't have an account)  they still get the same msg: `You will receive an email within the next few minutes. It contains instructions for changing your password.`

TheDudeWithTheThing:
well, that's more of a security "feature" so people can't guess emails

markottaviani:
Yeah, but does it really help?  I think in our situation it just causes
confusion.    The existing password reset system tells you if you entered
your email address wrong.

If we want to do some request throttling down the road to avoid some
hacking attempts I’m all  for it.

TheDudeWithTheThing:
h4x!

That's just what the gem is doing, I probably should have done some kind of audit to see how everything currently works first. I definitely didn't intend on changing anything too drastically. I did remove the ability to change your password in the profile page but now that I have a better grasp on wtf clearance is doing I can put that back I think.

TheDudeWithTheThing:
closing in favor of https://github.com/ophrescue/RescueRails/pull/875

PR #778, opened 2018-05-20
Users prefer dog manager to have no statuses set by default

markottaviani:
Had multiple feedback requests to adjust Dog Manager default view to not have any statuses set by default.


lazylester:
does this mean that there will be about 7000 dogs retrieved in the default manager view (with pagination, of course)? that's 250 pages of pagination1

markottaviani:
Yeah, big query that no one cares about.  We've had similar queries as defaults in the past and it hasn't been an issue, I'll make sure to keep an eye on the performance metrics in case anything gets weird.

PR #780, opened 2018-05-25
fix occasional test error due to custom rspec matcher bug

lazylester:
no production code changed, just rspec matcher and an rspec test added to confirm that the matcher is working as expected

### DIFF
--- a/spec/models/dog_search_spec.rb
+++ b/spec/models/dog_search_spec.rb
@@ -73,24 +73,28 @@ describe DogSearch do
       end
 
       context 'sort by name, confirm that rspec matcher anomaly is fixed' do
+        # selection of names is to trigger TravisCI sorting difference from development and production
         let!(:davinci){ create(:dog, name: "Da Vinci") }
         let!(:dallas){ create(:dog, name: "Dallas") }
         let(:params){ {sort: 'name', direction: 'asc'} }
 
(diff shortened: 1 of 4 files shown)


lazylester:
aargh travis failed on the very bug I fixed!  please standby for a fix to the fix.

markottaviani:
Which existing rspec test is this supposed to help out with?

lazylester:
spec/models/dog_search_spec.rb line 62 "searching for dogs by name with an asc sort".

spec/models/dog_search_spec.rb
```diff
@@ -71,6 +72,32 @@
...
+          else
+            expect(results.map(&:name)).to be_sorted(:ascending)
+          end
+        end
+      end
+
+      context 'confirm postgres compares strings as expected' do
+        let(:bool){ ActiveRecord::Base.connection.execute("select 'Da Vinci' < 'Dallas' as bool").first["bool"]}
+        it "really should return true. when this test fails on TravisCI, we'll know that the postgres sorting has changed" do
+          if ActiveRecord::Base.connection.current_database == "travis_ci_test"
```
markottaviani: Since this tests fails for me too, might need to adjust this to check for Ubuntu 14 instead vs Mac OS.

PR #783, opened 2018-05-29
fixes issue where flash persists to the next request

lazylester:
I just noticed that the flash refactor I did previously was allowing the "dog unavailable" message to persist onto the next request.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ coverage
 .powenv
 .vscode/
 .byebug_history
+site_notes.txt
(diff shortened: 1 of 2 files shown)


markottaviani:
Haven’t looked at the code yet, but the intended behavior is that the big
red banner that says “this dog isn’t available for adoption” should always
display when a public user of the site views a dog that’s already been
adopted.

We’ve had issues where someone finds a link to an already adopted dog,
misses the tiny status field that said “adopted”, applied for the dog and
then get mad when it was adopted 3 years ago.

markottaviani:
Great thanks!

PR #798, opened 2018-06-07
most recent photos were not being selected for inclusion

lazylester:
I noticed what appeared to be a bug in the rake task for adoptapet sync. The intention _appears_ to be that the urls of the four most recent photos should be included in the uploaded csv.

However, by inspection of the code, this was not happening, and  a test (see spec/tasks/adoptapet_sync_spec.rb) confirmed that it was not.

### DIFF
--- a/lib/tasks/adoptapet_sync.rake
+++ b/lib/tasks/adoptapet_sync.rake
@@ -20,9 +20,7 @@ namespace :adoptapet_sync do
       puts Time.now.strftime("%m/%d/%Y %H:%M")+ " Adoptapet #{state} Export Start"
 
       dogs = Dog.joins(:foster).where(
-        { status: ["adoptable",
-          "adoption pending",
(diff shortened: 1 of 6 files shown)


lazylester:
how is it possible... tests were added and coverage went down, with no change in the number of lines of code?

lazylester:
@iantalarico  added "in progress", but shouldn't the label be "code review"? Maybe my understanding of "in progress" is not correct?

iantalarico:
@lazylester This is my bad I started this a long time back but never got far.  Sorry!

lazylester:
@iantalarico I was not aware of any previous work on this, what were you doing?

iantalarico:
@lazylester oh wow never mind.  I thought this was an issue.  I have no idea how i added that label just now... Please ignore.  Sorry for the confusion...

PR #802, opened 2018-06-08
[WIP] draft for style review

lazylester:
In this PR I add Bootstrap 4.1 to all the views under (what was previously known as) DogsController. It's now split into two controllers. Code review is not recommended, just look at the views, on your laptops and on your phones.

At this point the code is not tested with rspec. When we have views that everyone is happy with, I'll go back and re-write the javascript and add thorough testing.

I have not been able to upload photos in my dev environment yet. So you may have to skip that activity. I need to get it working in my dev environment first.

You will need a populated database of course, so `rake rescue_rails:populate_all`. I have found it helpful to have real photos, so for this review, there's a list of photo urls I scraped from the website, and you can attach them to the populated dogs table with `rake rescue_rails:populate_photos`. In my experience it gives a more useful development environment to have real pics vs. fake pics.

I'll let you discover the page features, rather than introduce them. So I'll see you on slack or github or email for questions and clarifications.

enjoy!


app/views/dogs_manager/_dog.html.erb
```diff
@@ -0,0 +1,70 @@
...
+    <div class="<%= 'spay_neuter_needed' unless dog.is_altered %>" 'data-toggle'='tooltip' 'data-placement'='bottom' title='Needs Spay/Neuter'></div>
+  </div>
+  <div class='col-4 d-md-none'>
+    <a class="<%= 'fa fa-lg fa-comment-o' unless dog.comments.empty? %>"
+          tabindex='0'
+          data-toggle='popover'
+          data-placement='left'
+          data-trigger="focus"
+          title="Most recent comment for <%= dog.name %>" 
+          data-content="<%= dog.comments.first.content %>" ></a>
```
markottaviani: Is this line supposed to be ` data-content="<%= dog.comments.first%>" ></a>` .  ? Hit an error

app/views/dogs_manager/_dog.html.erb
```diff
@@ -0,0 +1,70 @@
...
+    <%- end %>
+  </div>
+  <div class='d-none d-md-block col-md-1'>
+    <a class="<%= 'fa fa-lg fa-comment-o' unless dog.comments.empty? %>"
+          tabindex='0'
+          data-toggle='popover'
+          data-placement='left'
+          data-trigger="focus"
+          title="Most recent comment for <%= dog.name %>" 
+          data-content="<%= dog.comments.first.content %>" ></a>
```
markottaviani: Is this line supposed to be ` data-content="<%= dog.comments.first%>" ></a>` .  ? Hit an error

markottaviani:
Nice!  Will probably need a little bit to poke around.  I'll add some comments as I see stuff.  Are dog photos in the gallery supposed to come up at random?

app/views/dogs_manager/_dog.html.erb
lazylester: hmmm... what error do you see? `dog.comments.first.content` is correct... it populates a bootstrap popover. And... obv... "works fine on _my_ machine"!

lazylester:
Yes @markottaviani there were supposed to be some real, random dog pics, scraped from the website, but looks as if this broke in the last few days and I forgot to fix it. I'll figure out what happened.

app/views/dogs_manager/_dog.html.erb
lazylester: I checked Chrome, FF and Safari, seems to work with those browsers.

lazylester: Oh I see the prob... the error occurs if there are no comments for a dog! quick fix coming...

markottaviani:
Great work with this!  

Here are a few quick feedback items of my own:
1. Consider adding a small thumbnail of the dog's photo to the Dog Manager Index View

2. It isn't obvious what the various icons mean (handcuffs, etc).  Consider adding alt/mouseover text on the index view, and also adding the icons to the show view next to the items they correspond to

3.  Bootstrap 2.3.2 regression:  The big center banner on the home page's buttons are set to the left instead of centered.

lazylester:
Thanks for the feedback, @markottaviani.

1. I can add a thumbnail on the dogs manager index view, but will have to skip it for the mobile version. Just not enough space.

2. Actually there _is_ a tooltip for the icons. It's a Bootstrap tooltip and even with the delay set to zero, it takes just long enough to appear, that one might easily miss the fact that it's available. I think I'll have to use a "home-made" tooltip iso Bootstrap, I'm sure faster response is quite achievable.

3. Strange about the home page regression. I'll check into it.

lazylester:
...thinking some more about the suggestion to consider adding thumbnails in the dog_manager#index page, there a couple of reasons not to do this:

1. It's hard! Compressing the data into two lines that would not break the format when the dog names, dog breeds, foster and adopter details are long was challenging. Adding another field without removing one each from line 1 and line 2 of each dog's display inevitably runs the risk of breaking things. We could go to 3 lines per dog, but then we add 50% more vertical page length.

2. All the mobile accommodations are done on the client side, with css. We do not, and should not, try to modify the view on the server side depending on the client device size. This means that the thumbnails will be downloaded to the mobile client, but not displayed. In a context in which bandwidth is a premium (with many more possible optimizations to follow this PR), it flies in the face of a mobile-optimized and performant page to add images, especially when they're never seen.

And finally, the dog's image is available just one click away, either by going to the 'show' page or else the gallery page (with, of course, the inevitable bandwidth penalty of the many images on the gallery page). 

And finally (really this time!) the current site does not show thumbnails in the manager index view. So we are not losing this feature in the BS41 upgrade.

With all that said, if we're willing to take the bandwidth hit on mobile, and the formatting penalty on larger screens, I'll happily add thumbnails!

markottaviani:
Fair enough! Don’t worry about it :)

lazylester:
@markottaviani this is now fully tested and ready for review by "The Rescue". Any chance of jumping to the front of the line, pushing to the staging server, and getting user feedback on this? If this is possible, it would be good, too, to update the staging db so that there aren't so many image anomalies e.g. in the dogs show page (from gallery view) most of the images show error messages, prob b/c they're gone from the production server? Oh, and if I could get an account on the staging server, I'd like to see what actual users are seeing.

markottaviani:
Sure thing, might take a little bit to get it out there given the changes.  But I should have some this evening to start.

lazylester:
@markottaviani I'm trying to understand the workflow for getting a deployment candidate pushed to staging for review. You mentioned that "given the changes" "it might take a little bit". Is the plan to pull the pr into the staging branch and deploy from there? Or can you deploy from (e.g.) a new branch called dogs_index_bs41? Or are you saying that you need to do a complete code review before deploying to staging (hopefully not, it could all change if they hate it)?

markottaviani:
Merging this into a feature branch within the main project, will make it easier to shuffle things arounds.

PR #807, opened 2018-06-19
refactor dogs#show page, move view logic into new carousel model

lazylester:
it's a refactoring extracted from the dogs_index_bs41 branch as an attempt to simplify the code review of that branch


markottaviani:
bringing this over to apply in the asset pipeline fix and then deploy out.

PR #811, opened 2018-06-21
Bootstrap 4.1 and new Dog Manager UI

markottaviani:
This feature branch will stay synced up with staging while we start working through review of new Bootstrap 4.1 and Dog Pages. 

Work items for this pull request should be attached to the Bootstrap 4.1 Milestone

### DIFF
--- a/app/assets/stylesheets/dogs_index/_manager_dog.scss
+++ b/app/assets/stylesheets/dogs_index/_manager_dog.scss
@@ -14,8 +14,19 @@
   }
 }
 
-
 #manager_list .dog.line-2 {
+  border-left: 1px solid #d4e4e4;
+  border-right: 1px solid #d4d4d4;
+  font-weight: 500;
+  padding-bottom: 6px;
+
+  .name {
+    font-weight: 700px;
+  }
+}
+
+
+#manager_list .dog.line-3 {
   border-left: 1px solid #d4d4d4;
   border-right: 1px solid #d4d4d4;
   border-bottom: 1px solid #d4d4d4;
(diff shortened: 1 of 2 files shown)


lazylester:
@markottaviani this branch now has adding and editing dogs fully Bootstrap 4.1 compatible and mobile-friendly. Not sure if you're getting any feedback regarding the dogs manager page on FB, or maybe you were waiting for add/edit dogs to be included. Anyway, I'd like now to get user feedback incorporated and wrap up this phase of Bootstrap 4.1.

markottaviani:
Great! Feedback has been very slim unfortunately.  Trying to track down
some of the power users to get some feedback.

lazylester:
yeah most ppl are just too busy for this kind of review. Personally, especially since it's not public-facing, I'd be inclined to deploy it and deal with any comments after. Then the review happens in real-world scenarios vs. just 'poking around'.

app/models/attachment.rb
```diff
@@ -34,7 +34,7 @@ class Attachment < ApplicationRecord
   belongs_to :updated_by_user, class_name: 'User'
 
   has_attached_file :attachment,
-            path: ':rails_root/public/system/attachments/:hash.:extension',
+            path: ":rails_root/public/system/#{Rails.env}/attachment/:hash.:extension",
```
markottaviani: That path value still gets used by S3, changing the target path breaks all existing image links.

lazylester: really? S3 uses this? Surely it's just for dev and test env, no? The rails_root/public/system path is in the Rails app, not the S3 server isn't it? Or else I'm misunderstanding something, (not the first time)

lazylester:
surprisingly tricky to make the file storage correctly configured for all environments, mainly b/c we are storing photos on S3 in the dog_photo directory (vs. convention 'photos'), combined with the fact we now have tests for file uploading, so we need an env-dependent configuration to facilitate cleanup.

Please try this fix on staging, and if it works ok we can have confidence in production. I don't know of a way to feature-test staging/production configs.

PR #819, opened 2018-06-23
Excel download of User Data

markottaviani:
Simple excel dump of user data to help management do cleanup of user accounts.  

Ideally the xls format would be blocked unless someone had admin permissions, instead of just a button hide.   So far my attempts at getting that to work in a `:before_action` have been futile.  I think that restriction is more of a nice to have than a must have.  

Regardless, if anyone has some suggestions, lmk.  Thanks!

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -89,7 +89,11 @@ class UsersController < ApplicationController
     @users = UserSearcher.search(params: params)
     respond_to do |format|
       format.html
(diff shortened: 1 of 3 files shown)


lazylester:
`respond_to do |format|
  if user.admin?
    format.xls { render... }
  else
    format.xls { head 'not_permitted' }
  end
end`

markottaviani:
Still want `active_user`s to be able to browse the directory, just don't want them easily walking off with the whole thing.

PR #826, opened 2018-07-05
reorganize S3 filesystem

lazylester:
To configure the staging and production servers for compatibility with this change, move the directory structure **/dog_photo/** to **/photos/**.


lazylester:
in addition to moving (copying prob better) `**/dog_photo/**` to `**/photos/**` it's also necessary to copy `**/event_photo/**` to `**/photos/**`.

TheDudeWithTheThing:
what should we do with this one?  Wouldn't we have to move all the current photos to the new structure?

markottaviani:
Yeah, would probably need to do an S3 shuffle.  I'll re-review where we're at with S3 hopefully in the next few days and see if the juice is worth the squeeze.

PR #853, opened 2018-07-18
fixes #852 Removes dog-status-definition page

### LINKED ISSUE #852: Remove dog_status_definition page Return 404
Per discussion in #841   I think `/dog_status_definition` page can be safely retired.  I probably wound up going to that page directly after seeing the translation error messages for the blue i on the Dog Gallery pages, saw the 406 error and went from there.

I can't find any reference to the page anywhere in the site.  I'm fine if we properly remove the page, we should just return a proper 404 instead.

lazylester:
Removes dog-status-definition page as it's no longer needed.


TheDudeWithTheThing:
@lazylester the title of this is remove dog-status but you removed status here, is that intended?

markottaviani:
Crap, good catch

Issue #838, opened 2018-07-16
Dog Manager index slow, big CPU usage spike post bs-4.1


markottaviani:
The default Dog Manager Index in production is taking about 9 seconds to render and is drawing a lot of CPU load as well.

markottaviani:
I think the first issue is that pagination is happening too late in the index method.  ActiveRecord goes and grabs everything plus all the includes before realizing it only really needed 30 dogs.  

```
  def index
    params[:filter_params] ||= {}
    @dog_filter = DogFilter.new search_params
    @dogs, @count, @filter_params = @dog_filter.filter
    for_page(params[:page])
  end
```

The first query that gets run is 
```
SELECT  "dogs".* FROM "dogs" ORDER BY tracking_id
```
No `LIMIT` or `OFFSET` so we wind up pulling back all the dogs and then proceeding to run a lot of complicated joins, creating a massive object of every single dog in the database.

If we instead include the `.paginate` as part of the search operation:
```
  def filter
    @dogs = Dog.search(search_params).merge(Dog.filter(filtering_params)).merge(Dog.has_flags(flag_params))
               .paginate(page: 1)
               .order( "#{sort_column} #{sort_direction}" )
               .includes(:adoptions, :adopters, :comments, :primary_breed, :secondary_breed, :foster)
    [@dogs, @dogs.length, params]
  end

```
We instead run `SELECT  "dogs".* FROM "dogs" ORDER BY tracking_id desc LIMIT 30 OFFSET 0` which is dramatically cheaper, and avoids the need to query and join nearly as much data.  

Now, what's the best way to get the page Param into the Dog Filter?
@TheDudeWithTheThing 
@lazylester

markottaviani:
PR #860 appears to have fixed the worst of the performance problems.

PR #860, opened 2018-07-19
Fix for dog index performance

markottaviani:
As per comments in #838 this sets pagination to run as part of the active record query.


markottaviani:
Lobbing this out now, but any retroactive suggestions are appreciated 👍

PR #871, opened 2018-07-23
raise errors on translation missing in dev & test env

lazylester:
catch these errors before deployment!


TheDudeWithTheThing:
this is a great gem too, you can run tests during the test suite to pick them up

https://github.com/glebm/i18n-tasks

PR #875, opened 2018-07-28
Adds clearance again, from master

TheDudeWithTheThing:
replaces PR https://github.com/ophrescue/RescueRails/pull/777

https://github.com/ophrescue/RescueRails/issues/580

- [ ] This PR: Replace current auth with clearance gem
- [ ] Switch to Bcrypt password strategy
- [ ] Sign in guards for logging IP, failures, etc...
- [ ] Clean Up remaining code from original implementation


markottaviani:
So I think we should add some kind of a password policy to this.  Right now it allows a user to set a one letter password.   😨

markottaviani:
Okay so for this PR, is "just" . _Replace current auth with clearance gem_  we'll break out the other items into new issues right?

-  Switch to Bcrypt password strategy
-  Sign in guards for logging IP, failures, etc...
-  Clean Up remaining code from original implementation

TheDudeWithTheThing:
Correct!

PR #886, opened 2018-08-06
conversion of events pages to Bootstrap 4.1

lazylester:


app/controllers/events_controller.rb
```diff
@@ -41,44 +41,27 @@
...
-    @events =
-      Event
-        .where("event_date < ?", Date.today)
-        .limit(30)
-        .order('event_date DESC')
-  end
-
-  def show
-    @event = Event.find(params[:id])
+    @scope = params[:scope] || "upcoming"
```
markottaviani: This got flagged on Hakiri.  
https://hakiri.io/github/ophrescue/RescueRails/master/5b60d169e706b7fc01fbec3b1edcc958627e44cb/warnings/dec3732cab1281

The only valid options for :scope are `upcoming` or `past` right?
Looking at https://brakemanscanner.org/docs/warning_types/dangerous_send/index.html  I think this can be changed to 
`    @scope = params[:scope] == "past" ? "past" : "upcoming"`

I at least tried it locally and the error didn't flag, and the tests passed 😄

lazylester: yes, the only valid options for :scope are `upcoming` and `past`. My thinking is that the definition of the route should preclude a user injecting some other value:
```
 get '/events/:scope', to: 'events#index', scope: /(past|upcoming)/, as: "scoped_events"
```
what do you think?

markottaviani: Ummm phone a friend @TheDudeWithTheThing

TheDudeWithTheThing: I mean you can never have too much sql injection prevention, why not both??

lazylester: why not both? because I'd like to think we're smarter than that! There is no benefit to applying the same filter twice, it certainly doesn't make the filter "extra strength".

The brakeman article cited by @markottaviani says that "using unfiltered user data ... is dangerous", "safer to whitelist". We do have a whitelist, it's in the route definition, _only_ "past" and "upcoming" are accepted as parameters. Hakiri failed to recognize this, not for the first time.

@TheDudeWithTheThing, this is not a sql injection vulnerability, the parameter is not used in a sql query.

If you see an actual security vulnerability here, vs. something that Hakiri is flagging, I'd love to hear the specific attack scenario.

TheDudeWithTheThing: I am in fact, not smarter than that.  

My only arguments, which are by no means great, are for developer friendliness and to shut Hakiri up.  At least in my experience I haven't used the route much to do this sort of thing so that'd be the last place I'd look. Can probably be marked as a me problem, not a code problem.  

Hakiri we can probably mark as a false positive and be done.

If we go with route, param or both we're covered so I'm good with this way.

TheDudeWithTheThing: Also, this makes it so you can't hit an event directly right?  Wouldn't going to `events/1` fail now or does rails translate show to `/event/1`, I forget

lazylester: `/events/1` still works. The routing algorithm looks for the first route that matches the incoming request url. So when `/events/:scope` with `:scope` matching the constraints `/(past|upcoming)/` fails to match, then the algorithm keeps looking and finds the `resources :events` route definition, which matches `/events/1`.

(note to self: should write a controller spec to confirm and document this behaviour)

TheDudeWithTheThing: please do!

markottaviani: Ideally we do want `/event/1` to work to support any direct linking to specific events.

lazylester: @markottaviani  I assume you mean `/events/1` should work, not `/event/1`. `/events/:id` *does* work, but it's not covered by test specs. AFAIK, however, it's not used anywhere, and it is not possible for anyone to get the url for a direct link to an event.

So, considering that nobody can know what the direct link is for an event, and that there are no links within the Rails app to single events, maybe it should be removed as it's unused?

markottaviani:
I’d rather we keep it, maybe toss a controller test in.

PR #896, opened 2018-08-14
fixes problems with galleria/carousel in IE11

lazylester:
by upgrading to the latest version of galleria.
Note that there is an incompatibility between this galleria version (1.5.7) and the rails asset pipeline that required a small change to the galleria.classic.js file.


lazylester:
this works on my local Windows10/IE11 stack, but still fails on Browserstack. I need to dig in and find out why.

lazylester:
ok to merge when appropriate. Fix of IE11/Galleria confirmed manually, pending fix of all IE11 issues

PR #902, opened 2018-08-25
bundle update

markottaviani:


markottaviani:
Test for location lookup failing, probably related to geocoder 1.5.0 and https://github.com/alexreisner/geocoder/issues/1258  Will need to dig into it

PR #915, opened 2018-08-28
Resources tests

lazylester:
adds baseline tests for staff resources (folders controller) in preparation for BS41 upgrade


markottaviani:
Conflict!

lazylester:
The Hakiri issue relates to rubyzip 1.2.1. The only place this is used is as a dependency of selenium-webdriver (once roo is removed with PR#923). The fix for this is very active in the rubyzip repo. I suggest we can live with this (as it's in test env, not production) and let the natural bundle upgrade process flush it out when the fix drops in the rubyzip repo.

PR #923, opened 2018-08-29
Removes unused roo gem which has security issue

TheDudeWithTheThing:
It looks like this is used in commented out code in the banned adopters controller.

@markottaviani do we still do banned adopter importing from spreadsheets?


markottaviani:
Nope, we don't, so yeah this can get dropped

PR #932, opened 2018-08-31
remove acts_as_list unused gem

lazylester:
Included only in the photo model, but apparently no longer used as the position attribute is managed by our own code.


lazylester:
see pr915 for discussion of hakiri warning

PR #949, opened 2018-09-08
Adds column last_login_at to user table

TheDudeWithTheThing:
There is currently lastlogin on the User table but I see nothing that has ever populated this column. To safely migrate away from that we'll create a new column in this PR and then use the column in the next PR.


markottaviani:
From a quick peek at the prd database, `lastlogin` database value was working prior to the clearance rollout.  

I think the old data in the old db field was certainly useful so I think as part of this db migration we should at least copy the data over from the old column.

TheDudeWithTheThing:
k going to just reuse the current column, the name annoys me since it's a date without the `_at` but I'll rename later, would rather get the functionality back quickly.

PR #950, opened 2018-09-08
Adds clearance sign in guard to log last log in at time

TheDudeWithTheThing:
Uses the current `lastlogin` column to log the time of a user's last login.  This is the first of possibly many sign in "guards" for clearance.


lazylester:
ha! I was just staring at the screen trying to figure out how to ask for tests without being a dick! And tests appeared. I just have to think it and it happens!

TheDudeWithTheThing:
haha, it's easy

> tests?

like I said, tests are very important to me, you'll get a happy path at least

PR #956, opened 2018-09-10
new rubocop todo for 0.59.0

markottaviani:
Yuck.  Maybe we can bake this in to travis to run only on modified files? Cut down on new stuff leaking in.


TheDudeWithTheThing:
well I'd rather lock the gem to a version until we get the violations down

PR #959, opened 2018-09-12
Stripe test

lazylester:
raise an exception if the donation_spec.rb is run without the stripe environment vars being set. Helps future devs!


markottaviani:
I think the schema.rb modifications are getting committed by accident?

lazylester:
AFAIK schema.rb has always been committed. This would be the "normal thing to do". Do you believe this to be a problem?

I'm not clear on the connection between schema.rb being committed and the test failure here, which seems to be due to configuration of env vars. Or am I totally confused?

markottaviani:
But this specific change doesn't involve changes to the schema.  Yet, this PR makes a LOT of changes to the schema. So I'm guessing something got goofed up in your local environment.  I had the same issue a few days ago.   You'll probably want to reset your local databases based on the schema in master.

lazylester:
@markottaviani nothing that I've done explains the changes to schema.rb. The changes came to me when I pulled the master branch from the repo, 6 days ago. I don't know how they got into the master branch of the repo. Could there have been an ActiveRecord update that made all these changes to the schema when a migration was done?

lazylester:
looks like there was a migration on 9/3 (that's the version that came in from the repo when I synced) that made all the changes to the schema.rb

markottaviani:
Running a

`git checkout master db/schema.rb`

should get you back on track

lazylester:
ok, but all that schema.rb stuff doesn't affect anything. The file is never used for anything, ever. And the build still fails at donation_spec.rb. It's possible that the change that was the whole reason for this PR doesn't work on TravisCI. But there's another later PR that is failing at the same place.

markottaviani:
Okay.. this is still failing because you're working off your own fork and thus don't have access to the credentials.  if we merge this to a feature branch with the main repo it should run just fine.

PR #963, opened 2018-09-12
Eliminate Folder description field

lazylester:
Here's the rationale:

**Motivation:**
This is in anticipation of the nested folders feature coming soon, by reducing the amount of information being shown on the Staff Resources page.

**Why it should be acceptable:**

- Out of 38 folders, only 13 have description fields.
- For the folders with description fields, the description field is mostly a superset of the name field. So the description field can be used as the folder name _instead of_ the name field, without losing information.

**Implementation**
A database migration is included in this PR which copies the description field (where it exists) to the name field, and then drops the description field of the folders model.


lazylester:
@markottaviani  the build failure symptoms are the same as I was seeing when the stripe credentials were not configured, so I'm assuming that there's still some issue with this. But I can't quite reconcile that theory with the fact that you (presumably) had the stripe feature building OK. Any insight?

Issue #947, opened 2018-09-07
Switch to Bcrypt password strategy


lazylester:
This can be done without requiring the user to reset their password:

1. add a second password field, `bcrypt_password`, to the User model
2. when a user first logs in, confirm their p/w using the current `encrypted_password` field, _and_ save their password bcrypt-encrypted in the new `bcrypt_password` field, and also set the `encrypted_password` field to nil
3. on subsequent logins, detect that the `encrypted_password` field is nil, so check against `bcrypt_password`

TheDudeWithTheThing:
I thought about this too but wanted to avoid hacking the clearance gem. Not only that but then the hack would live on forever as we have no real time frame of when all the passwords are migrated over.  I'd rather go clean break, it's not 100% unheard of to be forced to reset your password.

TheDudeWithTheThing:
> Hacking the clearance gem is as simple as overwriting a method in (e.g.) our /lib/clearance/xyz.rb file... no need to change the clearance code.

I agree, but then this code lives forever, unless we time box it. You have 2 months to change your password.

> but my email comes in on my computer, which is at home

I have to imagine 99% of the people doing this on their phone have their email hooked up to their phone, judging by the comments made on FB.

The trade offs here are, make it painful for our users for 5min or make it painful for us developers for the next few months.

lazylester:
What is the pain that developers suffer for the next few months? Maybe I'm missing something. It all happens without developer involvement. Nothing to think about until some point maybe a year in the future when all active users have switched to bcrypt passwords.

It's not "painful for the user for 5min" if the user is at an adoption event, trying to log in on someone else's computer, and they don't have access to their email account.

It can be made seamless and invisible to the user, so why not do that?

lazylester:
looks as if Thoughtbot has implemented a very [similar migration strategy](https://github.com/thoughtbot/clearance/issues/140) to the one I'm suggesting. It appears to be baked in to the Clearance code.

Instead of a new `bcrypt_password` column, they are able to detect whether the encrypted password string was produced by SHA1 or Bcrypt.

Thoughtbot, too, regards forcing users to reset passwords as undesirable when a seamless migration is possible.

TheDudeWithTheThing:
The pain comes when clearance has a security vulnerability in it and we have to upgrade which breaks whatever hack we have there. The main reason to move to this gem is so we don't have to think about login anymore. Having code in there that hacks / fights the gem doesn't get us there.

[...]

PR #969, opened 2018-09-15
Adds column woof_bark_password to users

TheDudeWithTheThing:
Step 1 here:
https://github.com/ophrescue/RescueRails/issues/947

### DIFF
--- a/db/migrate/20180915205719_add_woof_bark_password_to_user.rb
+++ b/db/migrate/20180915205719_add_woof_bark_password_to_user.rb
@@ -1,0 +1,5 @@
+class AddWoofBarkPasswordToUser < ActiveRecord::Migration[5.2]
+  def change
+    add_column :users, :woof_bark_password, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,30 +10,30 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_09_12_034937) do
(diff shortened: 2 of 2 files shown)


TheDudeWithTheThing:
what version of postgres are we on?  I'm assuming the removal of limits from the schema.rb are ok since we aren't doing a `rake db:schema:load` anywhere.

lazylester:
I note that the build failure is in a place where we've seen occasional failures before. I need to look into why this is happening.

TheDudeWithTheThing:
I’m not forcing a reset, you didn’t read my plan in the issue did you?

lazylester:
no, @TheDudeWithTheThing , I didn't read your plan. I didn't know it was there.

But now I've read it, thanks for the heads-up.

I don't really understand the rationale for the approach, but that doesn't matter, no explanation required. As long as it's all seamless to the users, I'm good with it. Last time we discussed it, you didn't see it as a big deal for the users to reset their passwords, and that's where we differ.

TheDudeWithTheThing:
I’m on board. The gist is with your plan we end up with a column named bcrypt_password which fights the gem. I’m just making a column up front for our woofbark password strategy so that bcrypt will only ever use the encrypted_password column.

lazylester:
Yes, the plan I suggested mentioned a second password field. But after suggesting that, I noticed that Clearance has a way to do it without a second field b/c they have a way to read the encrypted password and know what algorithm was used. I learned this after I had suggested a second field, as you mention.

PR #976, opened 2018-09-18
Adds encrypting password to bcrypt on successful login

TheDudeWithTheThing:
On successful login will resave password as the bcrypt encrypted version.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -16,6 +16,5 @@ User.destroy_all
           country: 'USA'
 }
 
- user = User.create!(@attr)
- user.toggle!(:admin)
-
+user = User.create!(@attr)
+user.toggle!(:admin)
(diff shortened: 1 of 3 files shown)


TheDudeWithTheThing:
The main reason is we are doing `Digest::SHA2` currently.

> I would suggest including a test that a user can log in with a bcrypt-encrypted password
This is every test except the one in this PR. When setting a password it will be encrypted via BCrypt.  I think the coverage dropped initially because every test was failing.

lazylester:
I have a lot of trouble navigating the Coveralls pages, so I could be quite wrong about this, but as far as I can tell, the bcrypt password validation is not covered by tests.

TheDudeWithTheThing:
@lazylester you might have caught it when they all failed.

https://coveralls.io/builds/19062586/source?filename=lib%2Fpassword_strategies%2Fwoof_bark.rb#L12

TheDudeWithTheThing:
I like that line 12 and 13 are exactly what my new test is testing, yet it shows up as untested.

lazylester:
coveralls has a credibility problem

PR #1000, opened 2018-09-29
Support for Monthly Donations

markottaviani:
This enables monthly donations via Stripe Billing Subscriptions.  

I still need to write a feature test for this, but otherwise I think it's in pretty good shape.  

Feedback appreciated!


TheDudeWithTheThing:
does stripe handle the monthly charge?  How does one cancel it if they want?

markottaviani:
This works via the “Stripe Billing” feature.  I actually didn’t find much in the way of clear cut guides for how to directly use Stripe Billing for monthly donations. So I probably have a blog post to write.

Cancellation gets handled via email to the rescue for the moment. I may be able to add some kind of a cancellation link via Stripe Billing emails.

PR #1001, opened 2018-09-29
Event clone fixes#997

lazylester:
much trickier than I expected!


lazylester:
build failing due to the stripe env vars issue. I'm still not clear how to avoid this in my workflow. AFAICT I am not able to create a branch in this repo to merge into (if I'm wrong about that, pls show me where I can do that). If you have any suggestion, @markottaviani I'd love to hear.

PR #1017, opened 2018-10-18
client-side validation of upload photos and documents fixes#1015

lazylester:
This will probably fail the build due to stripe limitations. I once knew, but now cannot figure out how to work around this limitation.


markottaviani:
cross merging to a new feature branch within the repo

PR #1148, opened 2019-03-26
Update account locking

markottaviani:
Fixed account locking and improves error messages.  Also will now immediately lock out a user with an active session.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -71,4 +71,14 @@ module ApplicationHelper
     render partial: 'shared/submit_or_cancel', locals: {f:f, return_path:return_path}
   end
 
+  def unlocked_user
+    if current_user.locked?
+      cookies.delete(:remember_token)
+      flash[:error] = 'Your account is locked.  You must contact Joanne@ophrescue.org to reactivate your account'
+      redirect_to(root_path)
+    else
+      return true
+    end
+  end
+
 end
--- a/lib/clearance/locked_account_guard.rb
+++ b/lib/clearance/locked_account_guard.rb
@@ -1,15 +1,16 @@
 module Clearance
-    class LockedAccountGuard < Clearance::SignInGuard
-      def call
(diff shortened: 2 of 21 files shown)


spec/factories/adopter.rb
```diff
@@ -23,7 +23,7 @@ def call_times
     training_email_sent { false }
     dog_reqs { Faker::Lorem.sentence }
     why_adopt { Faker::Lorem.sentence }
-    dog_name { Faker::Dog.name }
+    dog_name { Faker::Creature::Dog.name }
```
TheDudeWithTheThing: haha `Creature`

lib/clearance/locked_account_guard.rb
```diff
@@ -0,0 +1,16 @@
...
+    def call
+      if locked?
+        failure("Your Account is Locked.  You must contact joanne@ophrescue.org to reactivate your account.")
+      else
+        next_guard
+      end
+    end
+
+    def locked?
+      return if current_user.nil?
```
TheDudeWithTheThing: don't forget you can also do `current_user&.locked?`

app/helpers/application_helper.rb
```diff
@@ -71,4 +71,14 @@ def submit_or_return_to(f,return_path)
     render partial: 'shared/submit_or_cancel', locals: {f:f, return_path:return_path}
   end
 
+  def unlocked_user
+    if current_user.locked?
```
TheDudeWithTheThing: is this to handle people logged in already?  You could just also wipe out the sessions to force a logout on everyone.

TheDudeWithTheThing:
retroactive  👍

app/helpers/application_helper.rb
markottaviani: Yes, that's the idea.  In general I think it will be a good to have so they can immediately lock someone if they're being a jerk

markottaviani:
Thanks!

PR #1149, opened 2019-03-28
fixing adopapet sync test

markottaviani:
Any idea what I'm missing here?

I made a change in 160eed302cf0cdecfc6ee5abcd147c985cc5f28a to the adoptapet sync.   Change makes the adoptapet sync send the 4 top ranked photos by position order instead of 4 most recent uploaded photos.  

I didn't update the test 👎 

Trying to make right on that, but can't seem to get this thing green.  Any idea what I'm missing?
```
Failure/Error: expect(csv_photos).to eq photo_urls

       expected: ["/system/test/photos/0c3e99740ee5a5cb5b6b3ea702e40cf86fcda327.jpeg", "/system/test/photos/4fb926fd55...4b0cebf79b9792ff3a21e389.jpeg", "/system/test/photos/9727826a3cc0fe0d4cc94b974c69a4786035bdf3.jpeg"]
            got: ["/system/test/photos/0c3e99740ee5a5cb5b6b3ea702e40cf86fcda327.jpeg", "/system/test/photos/4fb926fd55...15be2af5a98040e0ce7f8af3.jpeg", "/system/test/photos/aefff0bcc9c49ce64b0cebf79b9792ff3a21e389.jpeg"]
```

### DIFF
--- a/spec/tasks/adoptapet_sync_spec.rb
+++ b/spec/tasks/adoptapet_sync_spec.rb
@@ -9,7 +9,7 @@ describe 'adoptapet_sync:export_upload', type: :task do
     let(:foster) { FactoryBot.create(:foster, region: "VA") }
     let!(:dog) { FactoryBot.create(:dog, status: 'adoptable', foster_id: foster.id, photos_attributes: [photo0,photo1,photo2,photo3,photo4].map(&:attributes)) }
     let!(:rake){ task.execute }
-    let(:recent_photos){ Dog.first.photos.sort_by{|photo| photo.updated_at}.reverse[0..3]}
+    let(:recent_photos){ Dog.first.photos.visible.sort_by{|photo| photo.position}}
     let(:photo_urls){ recent_photos.map{|photo| photo.photo.url(:large)} }
     let(:csv) { CSV.read( '/tmp/adoptapet/pets_VA.csv' ) }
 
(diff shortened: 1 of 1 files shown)


spec/tasks/adoptapet_sync_spec.rb
```diff
@@ -9,15 +9,15 @@
...
+    let(:recent_photos){ Dog.first.photos.visible.sort_by{|photo| photo.position}}
     let(:photo_urls){ recent_photos.map{|photo| photo.photo.url(:large)} }
     let(:csv) { CSV.read( '/tmp/adoptapet/pets_VA.csv' ) }
 
     it "should include most recent four photos for a dog" do
       csv_photos = csv.first[16..19]
       # just confirming recent_photos are the ones we want to upload
       # expect most recent first
-      expect(recent_photos.map(&:updated_at)).to be_sorted(:descending)
+      expect(recent_photos.map(&:updated_at)).to be_sorted(:ascending)
```
MatthewDrouin: ascending means that earlier values precede later ones e.g. 2018/1/1 will sort ahead of 2018/1/2 so you're not going to get `should include most recent four photos for a dog` you're going to get `should include oldest four photos for a dog with the oldest one coming first`

Not sure if that is your exact problem or not but your `it/do` does not match your code

markottaviani: Hey Matt.

Apologies I need to rewrite the spec description still.


The app code was changed so that instead of sorting for most recent photos, it should instead look at the `position` which is just an integer that sets the order of the photos.  We just want the photos in positions 1 thru 4.  

App code works right, just can’t seem to get the test right.

MatthewDrouin: No worries. Just trying to help ;)

MatthewDrouin: Should touch sort by position instead of updated_at?

markottaviani: Facepalm

Thanks!  Looks like it was that and a missing `[0..3]` to trim down the photo set.

Issue #996, opened 2018-09-27
Event Page should support Google Rich Results


markottaviani:
https://developers.google.com/search/docs/data-types/event

muhannadnouri:
I would like to work on that if that's okay. Let me know!

markottaviani:
@muhannadnouri I think you may have had a comment on this one already?  This would be a great initial contribution if you're still interested.

muhannadnouri:
I'll look into implementing this feature for an initial contribution!

Jngai:
I can work on this.

Jngai:
#1296

PR #1296, opened 2019-09-10
event structured data google

Jngai:
issue #996 

I checked the event's show page structured data with google structured data tool. 

Screen Shot 2019-09-10 at 3 08 06 PM


Jngai:
@markottaviani can you review my changes and help me out with the test? The tests hung up over package install issues. I am lacking a stripe env variable locally. If I have that then my tests should go green locally. thank you

PR #1298, opened 2019-09-11
Redo staff profile v1

Jngai:


Jngai:
@markottaviani can you check this pr and the test too. Getting the same error as the other pr. thank you.

PR #1185, opened 2019-05-01
Add birth date to adoption app

nickcianci:
Links to #710

PR #1401, opened 2020-02-05
Add birth date to adoption app (#1185)

markottaviani:
* Add birth date and adopter age to adoption app
* Add birth date and adopter age to adopt app mailer and add mailer preview
* Fix adopter age in adoption app and add tests
* Fix comment in mailer preview and make adoption age plural in all cases except for 1

(pulled into repo so that TravisCI can run)

### DIFF
--- a/spec/models/adoption_app_spec.rb
+++ b/spec/models/adoption_app_spec.rb
@@ -39,6 +39,7 @@
 #  shot_flea_tick            :boolean
 #  verify_home_auth          :boolean          default(FALSE)
 #  has_family_under_18       :boolean
+#  birth_date                :date
 #
 
 require 'rails_helper'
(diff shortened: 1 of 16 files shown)


spec/models/adoption_app_spec.rb
```diff
@@ -56,4 +57,36 @@
...
+      expect(adoption_app.adopter_age).to eq("10 years")
+    end
+
+    it 'returns 9 years when birth_date is 1 month before 10 years from current date' do
+      adoption_app.birth_date = (10.years.ago + 1.month).to_date
+
+      expect(adoption_app.adopter_age).to eq("9 years")
+    end
+
+    it 'returns 9 years when birth_date is 1 day before 10 years from current date' do
```
markottaviani: @nickcianci turns out this test will fail on late night CI runs :)    If UTC date and ET date are different this test will fail.  Given the org is fully Eastern Time based, it's probably best to honor local timezone of the app for determining age.  

I've added a quick time helper to consistently break the test which should make debugging a bit easier.  You should be able to push / pull directly from this feature branch.

nickcianci: I'm not able to replicate. Isn't it already honoring local timezone? I'm not honoring local timezone by using `10.years.ago + 1.day`?

markottaviani: Not sure, are you able to see https://travis-ci.org/ophrescue/RescueRails/jobs/647725489#L1778  ?

nickcianci: yes, can we print out the current datetime from the build server when running the test?

markottaviani: Sure can, just pushed some `puts`

nickcianci: Oh, I think I'm supposed to used Date.current instead of Date.today maybe.

nickcianci: https://stackoverflow.com/questions/12748626/when-should-datetime-now-utc-vs-time-current-utc-be-used-in-rails

markottaviani: Yeah I think that's it
```
Date.current is 2020-01-29
Date.today is 2020-01-30
```

markottaviani: Fix pushed

TheDudeWithTheThing: Timecop is great for this stuff too

PR #1476, opened 2020-04-03
allow dogs to be hidden from gallery

markottaviani:


TheDudeWithTheThing:
I thought we did this already

markottaviani:
We only hid dogs based on certain statuses, this allows them to just always hide specific dogs regardless of status.  
Expected to help with the mobs of people putting in applications for dogs.

Issue #849, opened 2018-07-17
Make dog search options bar more prominent


markottaviani:
>  I was wondering if it is possible to get a search bar back. For adoption coordinators it’s really time consuming to have to click through the filters.  When we do litters we have to look up 6-8 dogs unfortunately the way it is currently isn’t so user friendly
-Kate Polits

lazylester:
What search sequence is being envisioned here?

For example, if I'm searching for 6-8 dogs by name, the current UI adds just one extra click per search. Hardly "time consuming".  Can we request clarification of the sequence of searches?

markottaviani:
I think the two things to adjust are:

1. Have a way to click back to the search results when on the
dog_manager/show page
(Which existed in the old view).

2. Move the search options out onto their own row above the filters,
similar to how it was in the old version.  The search options checkbox
(name/id/breed/microchip) can probably get turned into a select box,
defaulting to name.

markottaviani:
Passing on a comment from a user that saw this github issue:

> To clarify his question we just want a search bar we don’t want to have to click a box to have to type something in. Before we just had a bar there ( similar to the url bar) and we just hit search. Now we have to find the drop down then click in the box. A lot of us use our phone to do things for OPH so with time it does get very time consuming.

lazylester:
Well, I'm not the best person to comment, since it was my judgement that the styling I designed was good! And that the single extra click is a small penalty for a tidy GUI. But...

If I were going to move the search box out of the dropdown. I would still keep the existing "search" icon and "clear" icon to trigger search and clear, and are only shown when there is a search string present, styled as [bootstrap button addons](http://getbootstrap.com/docs/4.1/components/input-group/#button-addons). I feel that the persistent bright blue bootstrap search button is a regression to a noisy UI design in which everything seems to compete for primacy.

The current compact style of search box can still be kept in its current position, vs. relocated above the other filter items. To me, this seems consistent with a rational visual information hierarchy.

I would suggest, too, looking at bootstrap's [buttons with dropdowns](http://getbootstrap.com/docs/4.1/components/input-group/#buttons-with-dropdowns) for the select button, because of its compactness.

It may be that entering a comma-separated list of names into the search field would provide a quick way to search for multiple dog names in a single search. Perhaps this would be a better way to make the commenter's searches more efficient.

[...]

PR #1517, opened 2020-05-05
Make dog search options bar more prominent

kateugenio:
#849 

Adds search bar on top of filter/sort card on Dog Manager index

### DIFF
--- a/app/views/dogs/manager/_manager.html.erb
+++ b/app/views/dogs/manager/_manager.html.erb
@@ -1,4 +1,4 @@
-<form class="mb-3" id="dog-search">
+<form class="my-3" id="dog-search">
   <div class="form-row">
     <div class="col-6">
       <input type="text" class="form-control" id='search' name='filter_params[search]' autocomplete='off' value=<%=@filter_params[:search]%> >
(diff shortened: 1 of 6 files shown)


app/views/dogs/manager/_manager.html.erb
```diff
@@ -27,7 +40,7 @@
   </div>
 </div>
 
-<div id="manager_dogs">
+<div id="manager_dogs" class="col-lg-12">
```
kateugenio, line 43: Minor update so that the list is aligned with the other rows (sorry my OCD kicked in).

app/views/dogs/manager/_manager.html.erb
```diff
@@ -1,12 +1,25 @@
...
+      <input type="text" class="form-control" id='search' name='filter_params[search]' autocomplete='off' value=<%=@filter_params[:search]%> >
+    </div>
+    <div class="col-2">
+      <%= select_tag :search_field_index, options_for_select(Dog::SEARCH_FIELDS.as_options.invert, selected: @filter_params[:search_field_index] || "name"), name: "filter_params[search_field_index]", class: "form-control" %>
+    </div>
+    <div class="col-4">
+      <button type="submit" id="search_button" class="btn btn-primary">Search</button>
+    </div>
+  </div>
+</form>
```
kateugenio, line 13: Not sure if we wanted to keep the `search_reset_icon` here so I didn't add it, let me know if that is preferred.

PR #1558, opened 2020-06-11
Feature/adoption fee

markottaviani:
Allows creation of Adoption Fee Invoices for adopters to easily pay their adoption fees online. 

Todo: Tests and review


app/controllers/invoices_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+                  :_method,
+                  :id,
+                  invoice:
+                  [
+                    :donation
+                  ]
+  end
+
+  def load_invoiceable
+    resource, id = request.path.split('/')[1, 2]
```
TheDudeWithTheThing, line 108: how come you aren't doing this with routes?  Also I'd check resource is a value that is allowed here, allowing user input to select the class doing a find makes me nervous, also anyway you can scope it to the current user's invoices?

app/models/invoice.rb
```diff
@@ -0,0 +1,72 @@
...
+                                       card: card_token
+
+    Stripe::Charge.create customer: customer.id,
+                          amount: amount * 100,
+                          description: 'Adoption Fee',
+                          currency: 'usd'
+  end
+
+  def open?
+    return true if self.status == 'open'
```
TheDudeWithTheThing, line 65: can shorten both of these to just be the condition

```suggestion
    self.status == 'open'
```

app/models/invoice.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def paid?
+    return true if self.status == 'paid'
```
TheDudeWithTheThing, line 69:
```suggestion
    self.status == 'paid'
```

app/controllers/invoices_controller.rb
```diff
@@ -0,0 +1,112 @@
...
+  def update
+    @invoice = Invoice.friendly.find(params[:id])
+    @invoice.card_token =  stripe_params["stripeToken"]
+
+    begin
+      @invoice.process_payment(@invoice.invoiceable.adopter.email)
+    rescue Stripe::CardError => e
+      flash[:error] = e.message
+    else
+      @invoice.donation = stripe_params["invoice"]["donation"]
```
TheDudeWithTheThing, line 60: get this wrapped in a transaction since you have 2 saves going on, I think you'd want `save!` too so something explodes if either goes awry, try that and if those are ok then send the email, I can see possibly sending an email to someone when something explodes

db/migrate/20200710000238_add_donation_to_inovices.rb
```diff
@@ -0,0 +1,5 @@
+class AddDonationToInovices < ActiveRecord::Migration[5.2]
+  def change
+    add_column :invoices, :donation, :integer
```
TheDudeWithTheThing, line 3: I'd make this a foreign key to the `donation_id`, which if null, means there was no donation. I somewhat don't like giving this a double meaning so you could even have a flag `has_donation` on invoices to indicate that and if that's true it better have a `donation_id` too.

app/controllers/invoices_controller.rb
markottaviani: An `Invoice` is polymorphic, and can belong to an `Adoption` or a `CatAdoption`, so this was needed to find the owning object.    As stolen from here: "The Ryan Bates Method" (RailsCast Dude)  https://codecabulary.launchacademy.com/learn-rails/polymorphic-associations

This could be waaay out of date way to handle it, but I hadn't had much luck finding better examples.

PR #1565, opened 2020-06-18
adopters set dog_or_cat on old apps

### LINKED ISSUE #1564: Adoption Coordinator should be able to set dog_or_cat on old applications
With the new `dog_or_cat` field now added, there are a lot of old applications where an adoption coordinator will want to be able to set this field on an existing application. 

Add to the Adopter show page an ability to set the value of the `dog_or_cat`  field if it is not already set.

ahutnick:
Creates a drop down menu if dog_or_cat not previously set to select Dog or Cat
Edited adoption_form.js to add autosave similar to Pet Selections
Fixed indentation on dog.html.erb

fixes #1564

### DIFF
--- a/app/views/adopters/_dog_choices.html.erb
+++ b/app/views/adopters/_dog_choices.html.erb
@@ -6,7 +6,7 @@
     <% else %>
       <td>
         <%= form_for @adopter, format: :json, html: { class: 'form-horizontal' } do |f| %>
-        <%= f.select :dog_or_cat, options_for_select([['Dog'], ['Cat']]) %>
+        <%= f.select :dog_or_cat, options_for_select([['Dog'], ['Cat']]), include_blank: true %>
         <div id="status-icon" class=""></div>
         <% end %>
       </td>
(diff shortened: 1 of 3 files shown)


app/views/adopters/_dog_choices.html.erb
```diff
@@ -1,7 +1,16 @@
 <table class="table table-striped table-bordered table-condensed">
   <tr>
     <th>Pet Type</th>
-    <td><%= @adopter.dog_or_cat%></td>
+    <% if @adopter.dog_or_cat? %>
+      <td><%= @adopter.dog_or_cat%></td>
+    <% else %>
+      <td>
+        <%= form_for @adopter, format: :json, html: { class: 'form-horizontal' } do |f| %>
+        <%= f.select :dog_or_cat, options_for_select([['Dog'], ['Cat']]) %>
```
markottaviani, line 9: The drop down should default to a blank value, otherwise it shows as "Dog" being already selected even though it isn't assigned.

`{ include_blank: true }`  
https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/select

ahutnick: Thanks!

Issue #1536, opened 2020-05-21
Handle second email address for adopters


markottaviani:
Adoption Coordinators that add a 2nd email address to the email value for an adoption application will cause automated emails and mailchimp subscriptions to not work correctly.

Suggested fix, add an Adoption Coordinator only field for  a 2nd email address.

Add validation rules to make sure it's impossible for AC's to add a 2nd email address to the email column.

kateugenio:
A few questions:

1. ..."_that add a 2nd email address to the email value_..", to clarify, is this if they click Edit and type in another email address in the input field? Example, if the original email value is `test1@test.com`, would they type something like: `test1@test.com, test2@test.com` in the field?

2. Is there a reason why the AC would need to update the applicant's email?

3. Just so I understand, why do AC's put in a 2nd email address? Is this a backup email for the applicant? Will it be used for as part of the automated emails?

Thanks!

markottaviani:
Great questions, thanks!

1. Yes that is the 'bad' behavior that the validation should check, catch and reject from the Adopter show Show page.

2. Yes, applications sometime mistype their email address or an applicant has decided they want to use a different email address, etc. 

3. The 2nd email address is often a spouse or other family member, really it acts as a backup or alternative contact.  So with that in mind, the 2nd email address _doesn't_ need any of the transactional emails or a mailchimp subscription.  However it should still work in the adopter search.   So if someone from the rescue gets an email from that person out of the blue a year from now, the volunteer can quickly search using the email address and hopefully find the right records!

PR #1589, opened 2020-07-07
Handle second email address for adopters

kateugenio:
#1536 

* Adds `secondary_email` to adopters
* Adds format validation to `email` and `secondary_email` in adopter model
* Refactors update action in AdoptersController to account for validation errors. To ensure error msgs were shown, the default template it was looking to render was `edit` (this is the way `respond_with` works on `update` actions), but there is no such template.


app/controllers/adopters_controller.rb
```diff
@@ -99,17 +99,21 @@ def create
   end
 
   def update
+    @title = @adopter.name
+    @adoption_app = @adopter.adoption_app
+    @adopter_users = User.where(edit_my_adopters: true).order("name")
```
TheDudeWithTheThing, line 104: this seems like it could return a ton of data, not sure the use case though, could it be hundreds?  thousands?

app/models/adopter_searcher.rb
```diff
@@ -37,7 +37,7 @@ def search
 
     if @params[:search]
       if email_search?
-        @adopters = @adopters.where('adopters.email ILIKE ?', "%#{@params[:search].strip}%")
+        @adopters = @adopters.where('adopters.email ILIKE ? OR adopters.secondary_email ILIKE ?', "%#{@params[:search].strip}%", "%#{@params[:search].strip}%")
```
TheDudeWithTheThing, line 40: strip once then use that value

```suggestion
        clean_email = @params[:search].strip
        @adopters = @adopters.where('adopters.email ILIKE ? OR adopters.secondary_email ILIKE ?', clean_email, clean_email)
```

app/controllers/adopters_controller.rb
kateugenio: Probably, I needed to set these instance vars in case we render the `show` template for the two cases below this. This is exactly what is set in the `show` action. Perhaps pagination could be another thing we can add?

markottaviani: This is the existing functionality we have for `show` and it's running pretty smoothly in production.

Issue #535, opened 2017-01-08
Break Down Stephs Application Enhancements


markottaviani:
> A couple suggestions coming from someone who stalks applications for specific things (namely the presence of kids, cats, or dogs). I spend a significant amount of time opening and going through applications in order to look for applicants to look for certain hard-to-adopt dogs. I could see the following suggestions helping many fosters as well. I do understand these updates aren't simple but things to consider.
> 1) When looking at applications the "first time dog owner" flag does not show if anything is put in the previous pets section. This counts if the only pets are cats. Also, its tough to tell if the dogs/cats are alive or not. Would like to suggest having check boxes to indicate if the applicant has dogs and/or cats and whether the animals are alive or not. This would allow the GUI to only show relevant questions (don't need feline questions if only dogs are in house)
> 2) On applications, we have an "Others in household" and age question - what about making this a series of checkboxes with values of: infants (under 1 yr), young child (2-7), older child (8-12), teenager (13-18), adults (19-65), senior (over 66). 
> 3) indicate type of home (apartment/condo, townhouse, single family, farm).
> 4) Indicate if there is a fenced yard (yes/no)

markottaviani:
These can now be done. You can be flexible w/ the UI

markottaviani:
Part 1 completed in #1578

PR #1592, opened 2020-07-11
Feature/household age selection #535

ahutnick:
Satisfies second bullet point on #535 

Adds :household_ages - an array storing string values for the age set
Question replaces :has_family_under_18 in the app
Displays results under Others in Household on applications show
Allows :household ages to be added on older apps 
Adds testing for :household_ages


markottaviani:
I think the age reference key should be moved to the left column: 

https://share.getcloudapp.com/yAuYqO22

PR #1628, opened 2020-08-01
Sarah/add question to foster profile/543

### LINKED ISSUE #543: Add to Foster Profile - question if they have any kids at all
> Do you think we can add the question for fosters if they have children? Currently we have children under five but I think it might be helpful to add if they have children period. If we can add that can we also make it one of the filters we can use for searching fosters?

sbuckingham89:
Resolves #543 

- Adds a separate question to the foster profile of whether they have children (in addition to the current question of whether or not they have children under five)

- Sets `has_children` to true for everyone that has selected `children_under_five`

- Adds additional filter for staff to use when searching for fosters


sbuckingham89:
> Hi Sarah, there were a bunch of changes to the user section layout (Upgraded to Bootstrap 4.x) probably after you first cloned the git repo a while ago. Can you please catch this up feature branch and resolve the conflicts?



> Hi Sarah, there were a bunch of changes to the user section layout (Upgraded to Bootstrap 4.x) probably after you first cloned the git repo a while ago. Can you please catch this up feature branch and resolve the conflicts?

Done!

PR #1678, opened 2020-09-03
Feature/training to each dog adopt #1659

### LINKED ISSUE #1659: TrainingMailer.free_training_notice should go to each Dog Adoptions
> enhancement requested via email Sub: 'Fwd: Free Training Session for your new dog'  from Laurie

TrainingMailer.free_training_notice  **currently** sends a notification email one time to an adopter, the first time their adoption application is set to  'adopted' or 'adptd sn pend'.  

Ref:  https://github.com/ophrescue/RescueRails/blob/8f13fd30e334d8c4b2b0305a9f0c3d00db4c6045/app/models/adopter.rb#L201-L207

OPH would like this behavior changed:

Only the adoption of dogs should trigger this email (not cats).
Each adoption of a dog should trigger this mail.  So someone that adopts 3 dogs should get 3 Training Emails (Will need to trigger off of the `Adoption.relation_type`)

ahutnick:
app/models/adopter.rb
- Training email now triggers off of changing relation_type to "adopted" && Dog adoptions only 
- Chimp check triggers off status change or relation_type change
- Followup email triggers off of changing relation_type to "adopted" && Dog adoptions only
spec/controllers/adopters_controller_spec.rb
- tests changed to prove emails trigger from relation_type change not status change
- added test to prove cat adoptions do not trigger training / followup emails
- added test to prove 2 emails sent per adoption 

closes #1659

### DIFF
--- a/app/models/adopter.rb
+++ b/app/models/adopter.rb
@@ -200,7 +200,7 @@ class Adopter < ApplicationRecord
 
   def training_email
     return unless status_changed?
-    if ((status == 'adopted') || (status == 'adptd sn pend')) && !training_email_sent?
+    if ((status == 'adopted') || (status == 'adptd sn pend')) && !training_email_sent? && dog_or_cat == "Dog"
       TrainingMailer.free_training_notice(id).deliver_later
       self.training_email_sent = true
     end
--- a/app/models/adopter.rb
+++ b/app/models/adopter.rb
@@ -156,7 +156,7 @@ class Adopter < ApplicationRecord
       }
     end
 
-    if ((status == 'adopted') || (status == 'adptd sn pend'))
+    if (status == 'adopted' || status == 'adptd sn pend') && adoptions.any?{ |a| a.relation_type == 'adopted' } && dog_or_cat == "Dog"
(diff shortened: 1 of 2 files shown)


markottaviani (approved):

app/models/adopter.rb
```diff
@@ -156,7 +156,7 @@ def chimp_subscribe
       }
     end
 
-    if ((status == 'adopted') || (status == 'adptd sn pend'))
+    if adoptions.any?{ |a| a.relation_type_changed? && a.relation_type == 'adopted' } && dog_or_cat == "Dog"
       completed_date = Time.now.strftime('%m/%d/%Y')
       AdopterFollowupMailer.one_week_followup(id).deliver_later(wait: 7.days)
```
markottaviani, line 161: The `AdopterFollowupMailer` probably not be triggered because the `status` of the `Adopter` and the `relation_type` of the `Adoption` each get set in the UI as a separate transaction.  So you'll need to asses the change of the relation_type directly when it is changed as its own separate action.   Bug filed as

PR #1707, opened 2020-09-29
#1696 make pet choices editable

### LINKED ISSUE #1696: Pet Choices section of Adopter Provile should be editable after submission
Only one of the three fields of "Pet choices" in the Adopter Show page is editable.  As shown in this screenshot:
https://share.getcloudapp.com/geuqvepA

The pet choices section should be modified to allow all three fields to be editable, using the same edit method used on the other tabs for example like this one:

https://share.getcloudapp.com/2NujYDmg

ahutnick:
- adds editable fields to each row in Pet Choices
- adds testing to ensure it works 
- adds Dog or Cat to Adopter Factory to prevent testing for nil value 

closes #1696

### DIFF
--- a/app/assets/javascripts/adoption_form.js
+++ b/app/assets/javascripts/adoption_form.js
@@ -26,10 +26,6 @@ $( function() {
       isEditing = !isEditing;
     });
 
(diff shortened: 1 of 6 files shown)


markottaviani (approved):
🥇

PR #1825, opened 2021-04-11
Feature/contracts

markottaviani:
Adds a custom action to `User` to create a release contract via REST API.


markottaviani:
fd223251a8c9e926c78cbacdc8311edbadfbebee needs some improvements.  This would call the REST api for each value, which is a total waste.

Need to get all the values for the given contract and then should be able to have them in the partial however they're needed.

app/controllers/users_controller.rb
```diff
@@ -141,6 +141,16 @@ def update
     end
   end
 
+  def create_release_contract
```
markottaviani, line 144: @TheDudeWithTheThing would this be the right way to handle this custom action? I'm unsure if i should have the actual method over in the model and call it instead.

PR #1875, opened 2021-06-18
user badge management

markottaviani:
This feature allows users to add and remove achievement badges from their profile by selecting from a list of available badges.

https://share.getcloudapp.com/o0ue6nmN

### DIFF
--- a/app/views/badges/_badge.html.erb
+++ b/app/views/badges/_badge.html.erb
@@ -4,6 +4,7 @@
         <%= badge.title %>
     </div>
     <div class="col-5 my-auto">
+        <%# TODO Get Rid of this N+1 Query %>
         <% if badge.users.where(id: current_user).exists? %>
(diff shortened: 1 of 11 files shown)


app/views/badges/_badge.html.erb
```diff
@@ -1,6 +1,13 @@
 <div class="row">
-    <div class="col-10">
-        <%= link_to(image_tag(badge.image.url(:thumb)),badge_path(badge)) %>
+    <div class="col-4">
+        <%= link_to(image_tag(badge.image.url(:thumb), alt: badge.title),badge_path(badge)) %>
         <%= badge.title %>
     </div>
+    <div class="col-5 my-auto">
+        <% if badge.users.where(id: current_user).exists? %>
```
markottaviani, line 7: @TheDudeWithTheThing  I need to provide a different link here for create or destroy depending on if the badge is attached to the `current_user` or not.  This is an N+1 query  as it wants to hit the database for each badge as it's displayed to see if it's attached.  

Any ideas on how to do this better?

